PR #74, opened 2026-05-04
feat: mcp

sergiofilhowz:
## Summary by CodeRabbit

* **New Features**
  * HTTP-only MCP bridge configurable via YAML (POST /mcp) with api_path, timeouts, hidden-prefix rules, and require_expose flag
  * --manifest output and simplified CLI: --config, --url, --manifest

* **Refactor**
  * Package/binary renamed to mcp and module surface streamlined; protocol reduced to core MCP v0.1 methods
  * Stdio transport removed in favor of HTTP bridge

* **Documentation**
  * README updated with quickstart, tool-name mangling, and reduced surface/behavior list

* **Tests**
  * Added comprehensive BDD suite and test helpers for core MCP behaviors


sergiofilhowz:
## CodeRabbit review pass

Worked through every actionable comment + the nitpick. Summary of what landed and what I'm deferring, with rationale.

### Addressed (10 of 11)

| # | Comment | Change |
|---|---|---|
| 2 + 6 | `api_path` leading-slash | Strip in `register_http_trigger` (defensive at the engine boundary) **and** in `load_config` (early-feedback warn at YAML load). New tests cover `/mcp` and `//mcp` cases. |
| 3 | `dispatch` swallows malformed frames | Extracted `validate_frame()` as a pure helper. Rejects non-object bodies, missing/non-string `method`, and `{"id":1}`-with-no-method as `-32600 INVALID_REQUEST` instead of falling through to the notification fast-path. 5 new unit tests. |
| 4 | `tools/call` bypassed `require_expose` | Added an *execution* guard: when `cfg.require_expose` is on, `tools/call` resolves the function via `engine::functions::list` and rejects with `tool_error` unless `metadata.mcp.expose == true`. The extra round trip is opt-in. |
| 5 | `register_all` swallowed trigger failures | Now returns `Result<(), IIIError>`. `main` logs `error!` (no longer claims `mcp ready`) on failure; the BDD harness logs a warn and keeps going (HTTP trigger isn't needed for direct `mcp::handler` invocation). |
| 7 | `__`-encoding is not bijective | Three-layer fix: (a) `debug_assert!` in `function_id_to_tool_name`, (b) `is_tool_name_ambiguous()` filters `__`-bearing IDs from `tools/list` with a warn log, (c) `tools/call` re-encodes the resolved function id and rejects with `tool_error` if the round-trip diverges. 2 new unit tests. |
| 8 | `engine.rs` `.ok()?` swallowed registration errors as skips | Switched to `.expect("failed to register MCP test workers")`. Connection failures still degrade to `None` for `--tags @pure`; *registration* failures are now loud. |
| 9 | Step-level soft-skip pattern | Kept the per-step `if world.iii.is_none() { return; }` (deliberate design for `--tags @pure` on engine-less dev hosts) but added a harness-level `III_ENGINE_REQUIRED=1` env gate in `bdd.rs` that panics at startup if the engine is unreachable. README documents the new env var for CI use. |
| 10 | `tools.rs` `required[0]` brittle | Changed to membership check over the `required` array. |
| 11 | (nitpick) `manifest` default_config drift | Now derives from `serde_json::to_value(McpConfig::default())`. Pulls in `require_expose` automatically and prevents future drift. New test asserts every `McpConfig` field is present in the manifest output. |

### Deferred (1 of 11) — **comment is based on a wrong premise**

| # | Comment | Reason |
|---|---|---|
| 1 | README says `prompts/list`/`prompts/get` are backed by the `skills` worker but the runtime calls `prompts::mcp-list` / `prompts::mcp-get` | The `prompts::mcp-list` and `prompts::mcp-get` functions **are registered by the skills worker** (see `skills/src/functions/prompts.rs:239,254`). They use the `prompts::` *namespace* but ship in the *skills* binary — there is no separate `prompts` worker in the workspace. The README's "Install `skills`" guidance is therefore correct, and the lower-table entries pointing to `prompts::mcp-*` describe the function IDs (also correct). No change needed. |

### Verification

- `cargo test --lib` — **48/48 pass** (8 of those are new: 5 for `validate_frame`, 2 for the `__` ambiguity check, 1 for manifest field completeness, 2 for `api_path` normalization).
- `cargo test --test manifest` — passes.
- `cargo clippy --all-targets -- -D warnings` — clean.
- `cargo test --test bdd` — 15/16 scenarios pass with my changes (same as without). The single failing scenario `tools/call without params returns -32602` was failing **before** my changes too — `iii.trigger("mcp::handler", body)` returns `Err` for a body with no `params` field, so the response never reaches the assertion. Worth a separate look (likely engine-side request-validation), but unrelated to this review pass.

### Files touched

```
mcp/README.md                |   5 +
mcp/src/config.rs            |  37 +-
mcp/src/functions/handler.rs | 258 +-
mcp/src/functions/mod.rs     |  42 +-
mcp/src/main.rs              |  23 +-
mcp/src/manifest.rs          |  37 +-
mcp/src/protocol.rs          |  44 +
mcp/tests/bdd.rs             |  32 +-
mcp/tests/common/engine.rs   |   9 +-
mcp/tests/common/workers.rs  |  12 +-
mcp/tests/steps/tools.rs     |   8 +-
```

PR #92, opened 2026-05-07
feat(sandbox-e2b): E2B microVM sandbox provider worker

rohitg00:
First member of the **sandbox worker family** + the family contract doc.

## What ships

`sandbox-e2b/` — narrow Rust iii worker that exposes [E2B](https://e2b.dev) microVM sandboxes via `sandbox::e2b::*` triggers. Live verified end-to-end against `api.e2b.dev`.

`sandbox-CONTEXT.md` — family-scoped CONTEXT.md captured from a grill-with-docs session. 10 resolved decisions covering terminology, ABI shape, idempotency contract, and evolution policy. See the file for details.

## Functions

| Function | Status |
|---|---|
| `sandbox::e2b::create` | wired against `POST /sandboxes` (forwards `idle_timeout_secs` as `timeout`) |
| `sandbox::e2b::exec` | stub — E2B exec runs through `/sandboxes/{id}/processes` SSE; streaming parser is a follow-up |
| `sandbox::e2b::stop` | wired; 404 treated as idempotent success |
| `sandbox::e2b::list` | wired; reconciles `in_flight` to upstream count |
| `sandbox::e2b::snapshot` | stub |
| `sandbox::e2b::expose_port` | stub |
| `sandbox::e2b::fs::read` / `fs::write` | stub |

Capabilities advertised: `["snapshot","expose_port","fs"]`.

## Live verification

Cycle run against `api.e2b.dev` with a real `E2B_API_KEY`:
- create → real sandbox booted
- list → reconciled `in_flight=1`, sandbox visible
- stop → `{}` success
- second stop on same id (sandbox already auto-killed) → `{}` (the idempotency fix this PR captures)
- E2B dashboard confirms `0 concurrent sandboxes` at end

## Tests

`tests/integration.rs` — 10 wiremock-backed cases:
- happy create + capability advertisement
- image allowlist rejection
- 5xx rollback (`in_flight` stays 0)
- 401 → S503, 429 → S500
- exec input validation
- 404 stop = idempotent success + counter release (regression test for the live bug)
- 5xx stop = S502
- list reconciles `in_flight` to upstream count
- list falls back to local counter when upstream is down

`cargo test`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check` all clean.

## ABI

Per `sandbox-CONTEXT.md`:
- Lifecycle floor (`create`, `exec`, `stop`, `list`) is the family-wide hard requirement
- Optional functions advertised via `capabilities[]`
[...]

PR #94, opened 2026-05-07
feat(sandbox-morph): Morph Cloud sandbox provider worker (Infinibranch)

rohitg00:
Member of the **sandbox worker family**. Read `sandbox-CONTEXT.md` (lands via #92) for family terminology and ABI contract.

## What ships

`sandbox-morph/` — narrow Rust iii worker wrapping [Morph Cloud](https://cloud.morph.so). Morph's wedge is **Infinibranch**: snapshot a running VM and branch it into N siblings with live process state preserved. This is the only worker in the family that registers `sandbox::*::branch`.

## Functions

REST endpoint paths verified via context7 against `cloud.morph.so/docs/api-reference/*`:

| Function | Wired |
|---|---|
| `sandbox::morph::create` | `POST /api/instance?snapshot_id=<image>` body `{metadata, ttl_seconds, ttl_action}` |
| `sandbox::morph::exec` | `POST /api/instance/{id}/exec` body `{command: [argv]}` |
| `sandbox::morph::stop` | `POST /api/instance/{id}/stop`; 404 / 409 = idempotent success |
| `sandbox::morph::list` | `GET /api/instance`; reconciles `in_flight` |
| `sandbox::morph::snapshot` | `POST /api/instance/{id}/snapshot` |
| `sandbox::morph::branch` | `POST /api/instance/{id}/branch?count=N` body `{snapshot_metadata, instance_metadata}` |
| `sandbox::morph::expose_port` | stub — Morph's `expose_http_service` lacks a clean REST equivalent today |
| `sandbox::morph::fs::*` | not registered — Morph fs ops go through SSH; not advertised as capability |

Capabilities advertised: `["branch","snapshot","expose_port"]`. No `fs`.

## Caller contract caveat

Morph requires a `snapshot_id` query parameter on create — there's no default rootfs to fall back on. The worker rejects empty `image` upfront with `WorkerError::BadInput` rather than letting the upstream return a confusing 4xx.

## Tests

15 wiremock-backed cases including the branch happy-path (returns `{sandbox_ids: [...]}` from Morph's `instances:[{id}, {id}]` response), `count: 0` rejection, snapshot-id extraction, 401 → S503, 5xx rollback, list reconciliation. Clippy clean (with `doc_markdown` allowed for URL-shaped doc comments).

## Live test status

Wired against documented shapes; not yet run end-to-end (no `MORPH_API_KEY` available during this PR's authoring). Pattern-matches the e2b/daytona fix cycle that was live-verified.

## Pin

`iii-sdk = "=0.11.6"`.


rohitg00:
Refactored to register only `sandbox::provider::morph::*`. Caller-facing `sandbox::*` now owned by the new sandbox router worker in #119. Direct invocation of `sandbox::provider::morph::*` stays supported.

Validation: `cargo fmt --check` + `cargo clippy -D warnings` + tests clean. Depends on #119 for end-to-end use via the router; this PR is independently mergeable if direct invocation is sufficient.

PR #96, opened 2026-05-07
feat(sandbox-modal): Modal sandbox provider worker (Python)

rohitg00:
Member of the **sandbox worker family**. Read \`sandbox-CONTEXT.md\` (lands via #92) for family terminology and ABI contract.

## What ships

\`sandbox-modal/\` — narrow Python iii worker wrapping [Modal](https://modal.com) sandboxes. Modal is gRPC-only (no public REST), so the worker imports the official \`modal\` Python SDK as its transport. The SDK is implementation detail; callers see only the canonical \`sandbox::modal::*\` ABI.

## Functions

| Function | Wired |
|---|---|
| \`sandbox::modal::create\` | \`modal.Sandbox.create(image=, app=, timeout=, cpu=, memory=)\` |
| \`sandbox::modal::exec\` | \`sb.exec(*argv)\` → \`{stdout, stderr, exit_code}\` |
| \`sandbox::modal::stop\` | \`sb.terminate()\`; idempotent on unknown sandbox_id (matches family contract) |
| \`sandbox::modal::list\` | local registry — Modal does not expose a server-side list-all primitive |
| \`sandbox::modal::snapshot\` | \`sb.snapshot_filesystem()\` — returns Modal Image \`object_id\` |
| \`sandbox::modal::expose_port\` | \`sb.tunnels()[port].url\`; surfaces clear S502 if port wasn't declared via \`encrypted_ports=[port]\` at create time |

Capabilities advertised: \`[\"snapshot\",\"expose_port\"]\`. No \`fs\` (Modal's FileIO API doesn't map cleanly to channel-based FS today).

## Implementation notes

- **Lazy import**: \`modal\` is imported inside each method via \`asyncio.to_thread\` so the worker module imports cleanly without Modal credentials. Tests mock \`ModalClient\` and never trigger the real import.
- **Handle registry**: an asyncio-lock-guarded dict maps \`sandbox_id\` (= \`modal.Sandbox.object_id\`) to the live Sandbox handle. Modal's SDK doesn't expose a server-side list, so we surface what the worker has tracked locally.
- **Idempotent stop**: an unknown sandbox_id, or a Modal \`NotFoundError\`, is treated as already-gone. Matches the family-wide stop contract.

## Auth

\`MODAL_TOKEN_ID\` + \`MODAL_TOKEN_SECRET\` — read by the \`modal\` package directly. Worker fails fast at first SDK call if missing.

## Tests

6 pytest cases mocking \`ModalClient\` (allowlist rejection, in_flight rollback, exec validation, idempotent stop on unknown id, list envelope, asyncio loop sanity). \`ruff check\` + \`ruff format\` clean.

## Live test status

Wired against documented signatures; not yet end-to-end verified (no Modal credentials during authoring).

## Pin

\`iii-sdk==0.11.6\`.

### DIFF
--- a/sandbox-modal/.gitignore
+++ b/sandbox-modal/.gitignore
@@ -1,0 +1,4 @@
+.venv/
+__pycache__/
+*.egg-info/
+*.pyc
--- a/sandbox-modal/Dockerfile
+++ b/sandbox-modal/Dockerfile
@@ -1,0 +1,12 @@
+FROM python:3.12-slim
+
+WORKDIR /app
+
+COPY pyproject.toml ./
+COPY src/ src/
+
+RUN pip install --no-cache-dir .
+
+ENV III_URL=ws://localhost:49134
+
+CMD ["sandbox-modal"]
--- a/sandbox-modal/README.md
+++ b/sandbox-modal/README.md
@@ -1,6 +1,6 @@
 # sandbox-modal
 
-Narrow iii worker that wraps [Modal](https://modal.com) sandboxes. Modal is gRPC-only — there is no public REST API — so this worker imports the official Modal Python SDK as its transport. The SDK is an implementation detail; callers see only the canonical `sandbox::modal::*` ABI.
+Narrow iii worker that wraps [Modal](https://modal.com) sandboxes. Modal is gRPC-only — there is no public REST API — so this worker imports the official Modal Python SDK as its transport. The SDK is an implementation detail; callers see only the canonical `sandbox::provider::modal::*` ABI.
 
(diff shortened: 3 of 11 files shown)


rohitg00:
Refactored to register only `sandbox::provider::modal::*`. Caller-facing `sandbox::*` now owned by the new sandbox router worker in #119. Direct invocation of `sandbox::provider::modal::*` stays supported.

Validation: `cargo fmt --check` + `cargo clippy -D warnings` + tests clean. Depends on #119 for end-to-end use via the router; this PR is independently mergeable if direct invocation is sufficient.

PR #103, opened 2026-05-07
feat(harness): consolidate shell-bash + shell-filesystem into shell worker

andersonleal:
## Summary

Replace the two old worker dependencies (`shell-bash`, `shell-filesystem`) with the consolidated **`shell` worker (v0.3.0)** that exposes `shell::*` (process lifecycle) and `shell::fs::*` (filesystem) under one binary. Migrates every harness call-site, palette regex, denylist string, test fixture, and doc reference to the new namespace.

**55 files changed**, +577 / −7917. Net deletions because the two old worker crates are removed.

## What changes

- **Removed:** `shell-bash/` and `shell-filesystem/` crates (39 files). Also drops `shell::bash::which`, `shell::bash::detect_clis`, `shell::filesystem::edit` — no equivalents in v0.3.0, no harness callers.
- **Worker manifest / build / docs:** `iii.worker.yaml` deps swap, `Makefile` + `scripts/demo.sh` `WORKERS` list swap, `--config` wiring for the shell worker (which refuses to boot unjailed by default), new demo-only `harness/shell-config.yaml` with `allow_unjailed: true`, `ARCHITECTURE.md` text + diagram + worker-count update.
- **Harness Rust:** `EXPECTED_WORKERS` uses `\"shell\"`; new test asserts the swap. New `harness/src/fs.rs` with:
  - `build_inline_envelope` (pure helper, TDD'd: UTF-8 happy path, truncation, binary marker fallback).
  - `harness::fs::read_inline` async driver — wraps `shell::fs::read`, drains its `StreamChannelRef` server-side under a 256 KiB cap, returns the legacy `{content:[{text}], details:{size, truncated, bytes_read}}` envelope so the FilesystemPanel preview keeps working without channel-awareness leaking into the web bridge.
- **Harness web (TS):** `FsLsResponse` for the flat `shell::fs::ls` shape; `FsEntry` widened with optional `is_dir`/`is_symlink`; FilesystemPanel `read` switches to `harness::fs::read_inline`; `palette.ts` `SENSITIVE_PATTERNS`, `App.tsx` `APPROVAL_REQUIRED`, and test fixtures all migrated to `shell::fs::*`.
- **Out of scope:** background-job functions (`shell::exec_bg`/`kill`/`status`/`list`) are exposed by the new worker but not surfaced through the harness palette/denylist/UI. Deferred until we design for them.

## Why

One worker, one binary, one config — eliminates the dual `shell-bash` / `shell-filesystem` boundary that didn't carry its weight (no separate enforcement surface, no separate lifecycle). The new worker also exposes background-job lifecycle that we can adopt in a follow-up without re-plumbing.

## Test plan

- [x] \`cargo test --manifest-path harness/Cargo.toml --lib\` — **32 pass** (incl. new \`expected_workers_includes_shell_consolidated\`).
- [x] \`cargo test --manifest-path harness/Cargo.toml --test integration\` — 2 pass (incl. \`expected_workers_matches_yaml_dependency_count\`).
- [x] \`cargo clippy --manifest-path harness/Cargo.toml --lib -- -D warnings\` — clean.
- [x] \`cargo build --manifest-path shell/Cargo.toml --release\` — clean release binary.
- [x] \`pnpm tsc --noEmit\` (web) — clean.
- [x] \`pnpm vitest run\` (web) — **121/121 pass**.
- [x] \`grep -rnE 'shell::(filesystem|bash)' harness\` — only intentional refs remain (test assertion messages + migration doc comments).
- [ ] **Manual smoke** (needs live engine + browser): \`make -C harness all\`, exercise FilesystemPanel ls + file preview, send one agent turn that calls \`shell::fs::ls\`. Write paths (\`shell::fs::write\`) intentionally not exercised — see R1 in design doc.

## Risks

- **Agent-driven \`shell::fs::write\`** may fail if the SDK tool-execution path doesn't auto-negotiate write channels for tool calls. The new \`shell::fs::write\` expects a write \`ContentRef\` in its payload (channel-streamed), whereas the old \`shell::filesystem::write\` accepted inline bytes. Verify during manual smoke; surface separately if broken.
- **Demo \`shell-config.yaml\` opts into \`allow_unjailed: true\`** so the FilesystemPanel can browse the developer's machine. Production deployments must pin \`fs.host_root\` instead. Documented in the file's leading comment.

## Notes for review

This branch is one squashed commit (\`f0f7eec\`). Original 19-commit history (Tasks 1–17 of the implementation plan + 2 cleanup commits + Cargo.lock regen) is preserved in local reflog if you'd prefer to review the granular sequence.


ytallo (approved):

sergiofilhowz (changes requested):
why are we deleting these two workers?

sergiofilhowz (approved):

ytallo (approved):

sergiofilhowz (approved):

PR #205, opened 2026-06-01
perf: harness telemetry/span-volume reductions

ytallo:
## Summary

Cut per-turn telemetry/span volume on the harness (streaming-delta coalescing, in-process event sequencing, no-op event suppression) plus a telemetry-import/deps housekeeping pass. All changes are observability-quality / cost reductions — no agent behavior change, no wire-contract or rendering change.

## Changes

- **Coalesce streaming deltas** (`turn-orchestrator/assistant-streaming/coalesce-deltas.ts`): merge consecutive same-type provider deltas (`text` / `thinking` / `functioncall`) into a single `message_update` on a ~60ms periodic flush. The console renders streaming text by appending `llm_event.delta`, so concatenating deltas is wire-identical — no console change. Cuts the per-token emits (and the spans/RPCs each generates) on streaming-heavy turns.
- **Localize the per-event sequence counter** (`turn-orchestrator/events.ts`): `nextSeq` no longer issues a persisted `state::update` per agent event; it uses an in-process per-session counter, with a random per-process epoch in the stream `item_id` so a restart can't collide with prior frames (`item_id` is opaque — never read by the console, stripped by the fanout, delivered in insertion order). Removes one engine state write (and its trigger evaluation) per event.
- **Skip `turn_state_changed` emit on no-op saves** (`turn-orchestrator/state-runtime/store.ts`): `persistRecord` only emits when the consumer-visible view (`toView`) changes — `transitionTo` bumps `updated_at_ms` on every write, which `toView` drops — removing a redundant `stream::set` (and its engine trigger-evaluation pass) per stale-skip / idempotent save. `state:created` is always emitted so a fresh session still seeds the console mirror.
- **Telemetry housekeeping**: migrate telemetry imports from `iii-sdk/telemetry` to `@iii-dev/observability` (0.16.1), `runtime/otel.ts` simplification, and the dependency bump.

## Behavior / compatibility

- No wire-contract or rendering change for streaming (delta coalescing concatenates; the console keys on `llm_event.type`).
- These changes (in-process counter, no-op suppression, coalescing) require a harness rebuild + restart to take effect at runtime.

## Test plan

- [x] Harness unit suite green (`pnpm test`, 1051 tests) + `pnpm typecheck`.
- [ ] Manual: confirm reduced span volume on a fresh trace (`engine::traces::tree`) — per-event `state::update` eliminated; coalesced streaming `message_update`s.

## Release Notes

* **New Features**
  * Added delta coalescing for streaming message updates, merging consecutive deltas for improved response streaming performance.

* **Improvements**
  * State change detection now only emits updates when meaningful changes occur, reducing unnecessary event emissions.

* **Chores**
  * Updated SDK dependency to latest version for enhanced observability support.

### DIFF
--- a/harness/package.json
+++ b/harness/package.json
@@ -58,10 +58,11 @@
     "iii-web": "./dist/web/main.js"
   },
   "dependencies": {
+    "@iii-dev/observability": "^0.16.1",
     "@opentelemetry/api": "^1.9.0",
     "chokidar": "^3.6.0",
     "commander": "^12.1.0",
-    "iii-sdk": "^0.12.0",
+    "iii-sdk": "^0.16.1",
     "pino": "^9.5.0",
     "uuid": "^11.0.3",
     "yaml": "^2.6.1",
--- a/harness/pnpm-lock.yaml
+++ b/harness/pnpm-lock.yaml
@@ -8,6 +8,9 @@ importers:
 
   .:
     dependencies:
+      '@iii-dev/observability':
+        specifier: ^0.16.1
+        version: 0.16.1
       '@opentelemetry/api':
         specifier: ^1.9.0
         version: 1.9.1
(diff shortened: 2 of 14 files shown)


ytallo:
Split out of the original combined PR: the Traces push live-refresh + page improvements now live in #206. This PR is harness telemetry/span-volume only.

andersonleal (approved):

PR #247, opened 2026-06-11
feat(provider-anthropic): Anthropic Messages API provider worker

ytallo:
## Summary

Stacked on #241 (`feat/llm-router-rs`) — first real provider implementing the router's provider protocol from `tech-specs/2026-06-agentic/llm-router.md`.

**provider-anthropic** — Anthropic Messages API provider behind [llm-router](https://github.com/iii-hq/workers/pull/241):

- `provider::anthropic::stream`: SSE → `AssistantMessageEvent` state machine with ping watchdog, abort-on-channel-close, and the shared error taxonomy
- Request assembly: api-key/OAuth auth modes, extended-thinking budget mapping with degradation warnings, prompt-cache markers (system prompt, tools tail, last stable assistant turn; `PROVIDER_ANTHROPIC_CACHE=0` kill switch)
- `provider::anthropic::refresh_models`: live `GET /v1/models` ∪ curated capability snapshot (context windows, output ceilings, thinking budgets, pricing), reconciled into the router catalog with no cold-catalog hole
- Declare-with-backoff registration, registration-token persistence in iii-state (with store retry so a transient state failure at first boot cannot strand the binding), re-declare on `router::ready`

**Streaming robustness:**

- Byte-level UTF-8 buffering across SSE chunks — multi-byte characters split across network reads decode correctly
- Truncated streams (EOF before `message_stop`) surface as error frames instead of a clean `done`
- Content-block routing keyed by the wire `index` field; unknown block types cannot corrupt neighboring blocks
- Cumulative-aware usage merge (no double-counting of `message_delta` totals)
- Structured Anthropic error-type classification, including mid-stream SSE `error` events and router bus errors (`registration_rejected` is not reported as transient)
- Empty-content message guards and redacted-thinking round-trip for thinking-enabled tool loops (adds the opaque `RedactedThinking` content-block variant to llm-router types — the only router-crate change in this PR, +4 lines)

**Docs:** consumer-facing README per the worker-readme guide, runnable `examples/stream_chat.rs` demo, provider cross-reference in the router README.

Agent exposure is locked down per `iii-permissions.yaml`: `provider::anthropic::*` is never agent-callable — the router invokes it worker-to-worker.

## Test plan

- [x] 77 unit tests passing (SSE state machine, wire adapters, thinking budgets, cache markers, curated merge, error taxonomy, config precedence)
[...]

PR #256, opened 2026-06-12
feat(provider-openai): OpenAI Chat Completions provider worker

ytallo:
## Summary

Standalone llm-router plugin worker for the OpenAI Chat Completions API, structurally mirroring `provider-anthropic/` (#247): token-gated registration with `router::ready` rebind, live model discovery reconciled through the router's single write path, SSE → `AssistantMessageEvent` streaming with ping watchdog and abort-on-channel-close, and typed upstream error classification.

### OpenAI-specific surfaces

- **Reasoning effort** (`reasoning.rs`): `thinking_level` maps to `reasoning_effort` per model family — each ladder branch traces to a documented 400 (o1 rejects the param entirely, chat-tuned variants take only the fixed default, pro tiers are high-only, xhigh is gpt-5.2+). Non-reasoning families omit the param and degrade with a warning.
- **Native structured output**: `response_format` passes through as `json_schema` (the Anthropic provider reports-and-ignores it).
- **Live-only catalog with generation filter**: the declaration carries no static models; a refresh fires right after registration. OpenAI's `GET /v1/models` returns bare ids (no capabilities, names, or limits — unlike Anthropic's API), so a local family-metadata table enriches known families and a **name denylist** drops known legacy generations (gpt-3*/gpt-4*/chatgpt*/o1/o3/o4; unrecognized future families pass through). Dated snapshots fold into their undated alias when both are live. Against the real API: 70 chat models → 23 current-generation rows.

### Behavior notes

- No key configured → catalog reconciles to empty and `router::provider::list` shows `configured: false`; chats fail loudly with a classified error.
- Env-var credential fallback (`OPENAI_API_KEY`) resolves in the llm-router process, same as every provider.

## Test plan

- [x] `cargo test` — 56 unit + 5 engine-backed integration (registration/token persistence, live-only catalog, filtered reconcile, chat with cost fill, upstream 401 classification, router-restart redeclare)
- [x] Live e2e against the local stack (engine + llm-router + harness + console): paste-a-key → auto-discovery (70 live models), routed chat on gpt-5.1, **function calling** (`engine::workers::list` round-trip), multi-turn recall, **reasoning effort** (thinking high on gpt-5.1 → correct bat-and-ball answer; degrade on gpt-4o-mini pre-filter), mid-stream abort, and the filtered catalog (23 rows, picker clean, chat on gpt-5 mini)

## Release Notes

* **New Features**
  * Added LLM Router worker enabling multi-provider LLM orchestration with unified chat, completion, and abort APIs.
  * Added OpenAI provider with streaming chat completions and live model discovery.
  * Added model catalog management with capability-based filtering.
  * Added provider registration system with credential and token management.
  * Added configuration schema composition for dynamic provider onboarding.

* **Documentation**
  * Added comprehensive documentation and technical specifications.

* **Tests**
  * Added integration test suites for router and provider workflows.


ytallo:
Duplicate of #252 — opened by mistake without checking for the existing PR. #252 is now retargeted to main with the same head.

PR #260, opened 2026-06-15
feat(approval-gate): standalone worker for human-in-the-loop function approval

ytallo:
## Summary

`approval-gate` is a new standalone iii worker that owns the policy and decision surface for human-in-the-loop function calls. It evaluates every dispatched call against a per-session permission model and either lets it `continue`, `deny`s it, or `hold`s it for a human — maintaining an ephemeral inbox of held calls and notifying consumers via two trigger types. It ships as its own crate (`approval-gate/`, lib `approval_gate` + bin `approval-gate`) so the gate can be deployed, versioned, and operated independently of the harness.

## What it does

The worker exposes three surfaces backed by one permission model:

- **The gate** (`approval::gate`) — a `pre_dispatch` hook the worker binds to itself at boot. For each call it takes one settings snapshot and evaluates, in order:
  1. `approval::*` / `configuration::*` target → **deny** (`human_only_function`, even under `full` — self-escalation defense), checked before the snapshot.
  2. mode `full` → continue.
  3. per-session `approved_always` hit → continue (honoured in **every** mode).
  4. mode `auto` + `always_allow` hit → continue (equality fast-path or `*`-glob).
  5. fall through to `policy::check_permissions`: `allow` → continue · `deny` → deny · `needs_approval` / unparseable reply → **hold** · transport failure / timeout → **deny** (`gate_unavailable` — fail closed, never an unattended hold).
- **The decision plane** — human/console-only RPCs: `approval::resolve` applies a decision to a held call; `set_mode`, `add_always_allow` / `remove_always_allow`, `approve_always`, `get_settings`, `clear_settings` manage per-session settings.
- **The pending inbox** — `approval::list_pending` (tenancy filters + cursor) and `approval::get_pending` read the ephemeral index of held calls.

**Modes:** per-session `manual` / `auto` / `full`, defaulting to the deployment's `default_mode`.

**Resolve flow:** `resolve { decision: "allow" }` calls `harness::function::resolve` with `action: "execute"` to release the call back through the dispatch pipeline; `decision: "deny"` calls it with `action: "deliver", is_error: true` and a rendered `DenialEnvelope` the model can read and adapt to. Resolve ordering is harness-first, then delete, then emit, so a crash never loses a decision (it leaks one record for the sweep to collect). An unknown `{ session_id, function_call_id }` returns `{ resolved: false }`, not an error — duplicate decisions race benignly.

**Settings:** lazily seeded. Reads (including the gate hot path) never write; the first mutation materializes the record from the current deployment defaults and writes it once. The gate reads tolerantly (a state outage degrades to defaults, which never widen access); mutations read strictly.

**Lifecycle / GC:** a pending record exists only while a call is held — no resolved-approval history is kept. Every record has explicit deletion paths (resolve, `harness::turn_completed`, `session::deleted`) plus a cron `approval::sweep` backstop that expires records past `expires_at`. All deletions funnel through a single `delete_with_gate` helper that uses an atomic null-set to decide exactly-once emission.

**Events:** `approval::pending_created` fires (off the hot path) when a call is held, carrying a self-sufficient, redacted record for notification copy; `approval::pending_resolved` fires exactly once per record when it leaves the inbox, with `outcome: allow | deny | timeout | aborted`. Bindings filter on `session_id` and subset-equality `metadata` for multi-tenant fan-out.

## Architecture

- **Standalone crate** with its own `[workspace]`: `cargo build` produces the `approval-gate` binary; `--manifest` emits the registry-publish manifest; `--config` / `--url` configure the YAML wiring and engine socket.
- **Direct-iii sibling calls.** Calls to other workers go through thin per-target wrapper modules — `state.rs` (`state::get/set/delete/list`), `harness.rs` (`harness::function::resolve`), `session.rs` (`session::get`) — each a small `iii.trigger()` wrapper. There is no bus/transport abstraction; handlers receive `Deps { iii, sink, defaults, cfg }` and call iii directly.
- **One function per file** under `src/functions/`, registered through a single typed helper in `functions/mod.rs` that maps `ApprovalError` to the wire `code: message` shape. Pure logic (evaluation order, redaction, denial rendering, settings math) lives in dedicated modules (`decision.rs`, `redact.rs`, `denial.rs`, `settings.rs`, `pending.rs`).
- **State surface:** two scopes — `approval_pending/<sid>/<cid>` (the inbox) and `approval_settings/<sid>` (per-session settings) — both with explicit deletion paths. `/` is the reserved key separator, so `session_id` / `function_call_id` are rejected if they contain it.
- **Boot order** (`main.rs`): register the two trigger types → register the 14 functions → best-effort bind the hook / configuration / `session::deleted` / `turn_completed` / cron triggers → register the `approval-gate` configuration entry and read defaults once. Bindings are best-effort: on an engine without those trigger types the worker still boots, serves its RPCs, and logs `trigger_type_not_found` for the absent bindings (restart to re-bind after the sibling lands).
- **Configuration:** deployment defaults (`default_mode`, `always_allow_seed`, `pending_timeout_ms`) live in the `approval-gate` configuration entry with reactive reload via `approval::on_config_change`; field-wise tolerant parsing never fails the gate open. `config.yaml` carries runtime wiring only (hook globs/budget, sweep cron `expression`, per-call timeouts).

## Public contract

A consumer (notably the integration PR stacking on this) depends on:

**Functions** (`approval::*`):
- Harness-driven: `gate`
- Human/console: `resolve`, `set_mode`, `add_always_allow`, `remove_always_allow`, `approve_always`, `get_settings`, `clear_settings`, `list_pending`, `get_pending`
- Internal trigger handlers (never call directly): `on_config_change`, `on_session_deleted`, `on_turn_completed`, `sweep`

**Trigger types:** `approval::pending_created`, `approval::pending_resolved` (binding config `{ session_id?, metadata? }`, `deny_unknown_fields`).

**Errors:** `code: message` with `approval/invalid_payload`, `approval/state_unavailable`, `approval/harness_unavailable`.

**Harness handoff** (`approval::resolve` → `harness::function::resolve`): allow ⇒ `{ session_id, turn_id, function_call_id, action: "execute" }`; deny ⇒ `{ ..., action: "deliver", is_error: true, content, details }`. The gate binds itself via the `harness::hook::pre_dispatch` trigger type. These harness surfaces are greenfield: the current harness does not implement them, so until the stacked integration PR lands the bindings are inert and the gate is exercised against in-process fakes (see Testing).

## Testing

- **Engine-backed tests** (`approval-gate/tests/integration.rs`): spawn a **real** iii engine via `src/testkit/engine.rs` with `configuration` + `iii-state` real, faking the not-yet-built siblings (`policy::check_permissions`, `harness::function::resolve`, `session::get`) in-process. They cover the hold→record→single-emit idempotency path, allow/deny resolution through the fake harness, sweep expiry with exactly-once `timeout` emission, reactive configuration reload, lazy settings seeding against real state, and full-stack human-only denial.
- **Handler unit suites** sit beside each `approval::*` function and drive the same production code path through a real engine via `testkit::engine`. Pure-logic modules (`decision`, `redact`, `denial`, `settings`, `types`, …) are tested with no engine.
- **Self-skip:** engine-backed suites resolve the engine from `III_ENGINE_BIN` or `iii` on PATH and self-skip (printing a notice) when neither is present, so they never fail in an engine-less CI lane — CI must provide an engine binary or that coverage silently drops.
- **Status:** `cargo check` / `cargo build` clean; 94/94 tests pass. One integration test can flake under the parallel multi-thread engine harness (timeout); green when run single-threaded (`--test-threads=1`).

## Reviewer notes

- **Start with** `architecture/integration.md` (the consumer-facing function + trigger contract and the harness handoff) and `architecture/internals.md` (evaluation order, the delete/emit gate, lazy seeding, the engine facts the code relies on).
- **Then** `src/functions/gate.rs` (control flow / fail-closed posture) and `src/functions/resolve.rs` (crash ordering, the execute/deliver envelope) — these define the load-bearing behavior.
- **Scrutinize:** the exactly-once emission gate in `src/pending.rs` (`delete_with_gate` — relies on `state::set` storing a literal null tombstone, hence the two-step set-null-then-delete); the fail-closed mapping in `policy.rs` (transport failure/timeout ⇒ deny, unparseable ⇒ hold); the tolerant-vs-strict settings reads in `settings.rs`; and the `human_only_function` prefix check running before the settings snapshot in `gate.rs`.
- **Note** the per-target iii wrappers (`state.rs`, `harness.rs`, `session.rs`) are intentionally thin and untyped at the boundary — the typed shapes live in `types.rs`.

## Summary by CodeRabbit

* **New Features**
  * Introduced the **approval-gate** worker with a fail-closed pre-dispatch hold/resolve workflow.
  * Added pending inbox APIs (**list/get pending**) plus **pending created/resolved** notifications.
  * Added per-session approval settings management, including **mode** and **always-allow/approve-always**, with cron-based expiry sweeping.
* **Documentation**
  * Added end-user and architecture docs describing decision flow, triggers, and operational setup.
* **Tests**
  * Added unit coverage and engine-backed integration tests for gate/resolve/sweep and configuration reload.
* **Chores**
  * Updated CI to install the required engine before running Rust tests.


ytallo:
Companion harness integration PR stacks on this branch: #261 (adopts this worker, retires the in-process gate). Merge this one first.

PR #261, opened 2026-06-15
feat(harness): adopt the standalone approval-gate worker, retire the in-process gate

ytallo:
## Summary
The harness no longer ships its own approval policy engine. It exposes a generic synchronous pre-dispatch hook point (`harness::hook::pre_dispatch`) and a release surface (`harness::function::resolve`), and delegates every allow / deny / hold decision to the standalone approval-gate worker that binds to that hook. All of the in-harness gate code — permission modes, allow-lists, YAML-rule evaluation, redaction, and the per-session settings functions — is deleted. From the agent's and user's perspective the gating behavior is unchanged; only its owner moved out of process.

## Stacked on
Targets `feat/approval-gate-worker`, not `main`. The standalone worker crate lives on that base branch; this PR is the harness-side integration only and must merge **after** the base.

## What changes
- **New hook contract** (`src/turn-orchestrator/hooks/`): `harness::hook::pre_dispatch` is registered as a trigger type. Any sibling worker binds to it (priority-ordered, glob-filterable by function id) and answers `{decision: "continue" | "deny" | "hold"}` synchronously per agent function call.
  - `hooks/types.ts` — wire schema for hook input/output and the strict per-binding config (`functions`, `priority`, `timeout_ms`, `on_error`).
  - `hooks/registry.ts` — subscriber registry; bindings rebuild on engine replay after a harness restart.
  - `hooks/chain.ts` — consults bound hooks in order; first `deny`/`hold` short-circuits, all-`continue` ⇒ allow. Transport failure follows the binding's `on_error` (`fail_closed` default denies with `gate_unavailable`; `fail_open` skips). Zero bindings ⇒ allow (a deployment with no gate is ungated).
  - `hooks/denial.ts` — denial-envelope shapes shared by the hook chain and the generic trigger-failure path; `denied_by` is now `permissions | user | hook | gate_unavailable`.
- **Dispatch chokepoint** (`agent-trigger.ts`): `dispatchWithHook` now takes a `DispatchContext` (`session_id`, `turn_id`, `step`, `metadata`) and consults `consultPreDispatch` instead of the old in-process `consultBefore`. A `hold` returns `{kind: 'pending', held_by, pending_timeout_ms}` so the batch can park the call.
- **Release / deliver surface** (`function-resolve.ts`, new): `harness::function::resolve` lets the hook owner settle a held call — `action: "execute"` releases it back through the normal execution path (released calls skip the hook chain), `action: "deliver"` answers it with supplied content/`is_error` without executing (user deny, sweep timeout). Decisions are persisted to a `function_resolutions` scope and the parked turn is woken on the turn-step queue; unknown/stale/duplicate calls return `{resolved: false}` (benign, never an error). A transient state-read outage surfaces as an error so the caller retries rather than losing the hold.
- **Awaiting-approval flow** (`function-awaiting-approval/`): the wake now consumes `function_resolutions` rows (read → apply → delete) rather than reacting to an `approvals`-scope state-write trigger. It re-scans until a pass settles nothing new (a released call can settle a sibling), then routes the batch to `function_execute` or `finalizeBatch`.
- **Terminal cleanup** (`turn-completed.ts`, new): `harness::turn_completed` trigger type fires on terminal turns (completed / cancelled / failed) so the gate can purge a turn's pending records instead of polling. Fired from `finalizeBatch` and on stale-turn takeover in `run-start.ts`; at-least-once, unordered, idempotent.
- **Turn identity** (`state.ts`): records carry a `turn_id` (generated in `newRecord`, backfilled for pre-existing records) threaded through the hook input, `function::resolve`, and `turn_completed` so siblings can key their state to one turn.
- **Abort** (`run-abort.ts`): clears parked `awaiting_approval` and best-effort deletes their unconsumed resolution rows so a decision racing the abort doesn't linger.
- **Registration** (`turn-orchestrator/register.ts`): registers the two trigger types first (siblings can bind immediately), plus `harness::function::resolve`.
- **Config** (`harness/register.ts`, `index.ts`, `harness/iii.worker.yaml`): the `approval-gate` worker is removed from the composite `WORKERS` array and its dependency drop, and the harness config entry no longer carries the permissions block tied to the in-process gate.

## Removed
- `src/approval-gate/**` — the entire in-process gate: `resolve.ts`, `denial.ts`, `redact.ts`, `schemas.ts`, `main.ts`, `iii.worker.yaml`, and `settings/**` (mode, always-allow add/remove, default-mode, human-only, store, register, etc.).
- `src/turn-orchestrator/hook.ts` — the old `consultBefore` approval consult (human-only self-escalation defense → settings snapshot → mode/allow-list → `policy::check_permissions`), replaced by `hooks/chain.ts`.
- `src/harness/permissions-config.ts` and its `registerHarnessConfigEntry` wiring.
- Corresponding tests: `tests/approval-gate/**`, `tests/turn-orchestrator/hook.test.ts`, `tests/turn-orchestrator/function-awaiting-approval-state-trigger.test.ts`, `tests/integration/mode-approval.e2e.test.ts`.

## Behavior
- **Before:** the harness evaluated approval itself — permission modes, the curated `always_allow` list, YAML policy rules, and redaction all ran in-process inside `consultBefore`, and the awaiting-approval wake fired off an `approvals`-scope state write.
- **After:** the harness only renders allow/deny/hold outcomes into function results; the full policy (modes, allow-lists, YAML rules, redaction) lives in the standalone approval-gate worker bound to `harness::hook::pre_dispatch`. Releases come back via `harness::function::resolve`; cleanup is driven by `harness::turn_completed`.
- **Unchanged for the end user:** allowed calls execute, denied calls return a `denied` result (still classified as an error and rendered as `[PERMISSION_DENIED]`), held calls park the turn until a human decides, and abort still unparks cleanly. A gate that crashes/times out fails closed by default. With the standalone gate bound, the agent-facing outcome is identical to before.

## Testing
- `pnpm run typecheck` (`tsc -b --noEmit`) → **pass**, clean compile.
- `pnpm run test` (`vitest run`) → **74 files, 835 tests, all passing**, including `function-resolve`, `function-awaiting-approval`, `hooks/chain`, `hooks/registry`, `turn-completed`, `hook-gate-flow.e2e`, `wire-parity`, and `parallel-approval.e2e`.
- `pnpm run lint` (`biome check`) → clean for the files in this change (`lint:fix` applied). The only remaining warnings are pre-existing `noTemplateCurlyInString` notices in an unrelated `llm-router` config test, untouched here.

## Reviewer notes
- Start at `hooks/chain.ts` (the deny/hold/fail-open/fail-closed decision matrix) and `function-resolve.ts` (the `execute` vs `deliver` settlement and the strict-vs-tolerant read contract — a state outage must surface as an error, not `{resolved: false}`).
- Scrutinize the **hold → resolve → wake** loop in `function-awaiting-approval/run.ts`: the re-scan-until-stable pass, the `executed`-guard against double execution, and the extra single-shot wake when a pass settles some-but-not-all siblings.
- Check the denial path in `hooks/denial.ts` + `agent-trigger.ts`: `gate_unavailable` on hook failure, and that `status: 'denied'` still flows through `isErrorResult` and `[PERMISSION_DENIED]` rendering.
- Confirm `turn_id` backfill in `state.ts` keeps in-flight turns valid across the deploy, and that `turn_completed` fan-out (incl. the stale-takeover emit in `run-start.ts`) is genuinely fire-and-forget / idempotent.
- Verify `register.ts` ordering (trigger types before FSM steps) so a sibling that boots early can bind without a gap.


ytallo:
Stacked on #260 — that worker PR is the base; review/merge it first. GitHub will retarget this PR to `main` automatically once #260 lands.

sergiofilhowz (approved):

PR #267, opened 2026-06-15
feat(iii-directory): migrate runtime config to the configuration worker

andersonleal:
## Summary

`iii-directory` was the last common Rust worker still loading its runtime config from a local `config.yaml` parsed once at boot. This migrates it to the engine-built-in **`configuration` worker** (entry id `iii-directory`), following the **`database`/`storage` pattern**: register a JSON Schema + seed at boot, read the authoritative (env-expanded) value via `configuration::get`, and bind a `configuration:updated` trigger for hot reload.

The change uses **full hot-reload parity** with `database`:

- **Tunable (hot-reload live):** `registry_url`, `download_timeout_ms`, `registry_cache_ttl_ms`, `filter_unregistered`
- **Topology (refused with a "restart required" log):** `skills_folder`, `local_skills_folder`, `auto_download` — these define on-disk roots and boot-time task wiring.

## What changed

- **`src/configuration.rs` (new)** — port of `database/src/configuration.rs`. `CONFIG_ID = "iii-directory"`, `CONFIG_FN_ID = "directory::on-config-change"`. A `SharedState` bundles the live config handle, shared cache-TTL cell, both caches, and the boot topology. `on_config_change` re-fetches the authoritative value (ignores the trigger payload — same security rationale as `database`), refuses topology changes, otherwise swaps the live config + updates TTL + clears both caches.
- **`src/config.rs`** — `SkillsConfig` derives `JsonSchema`; adds `json_schema`/`from_json`/`to_json`/`from_yaml`/`from_file`/`topology`/`into_shared` and the `SharedConfig` alias (`Arc<ArcSwap<SkillsConfig>>`).
- **Live-config refactor** — every handler in `functions/{mod,skills,registry,prompts,download}.rs` snapshots the live config per call via `load_full()` (lock-free). `RegistryCache` and `RegisteredWorkersCache` read a shared `Arc<AtomicU64>` TTL so `registry_cache_ttl_ms` also hot-reloads.
- **`src/main.rs`** — `--config` is now an optional seed; boot reorders to connect → `register_config(seed)` → `fetch_config` → build state/caches → register functions → `register_config_trigger`.
- **Docs/seed** — `config.yaml` → `config.yaml.example` (seed header), README "Configuration" section rewritten (tunable vs topology), `manifest` `default_config` completed.
- Adds the small `arc-swap` dependency (lock-free per-call reads; one-line change at each handler site).

## Verification

- `cargo build`, `cargo clippy --all-targets` (clean), `cargo fmt --check` (clean)
- `cargo test --lib` — **249 passed**
- `cargo test --no-run` — all test targets compile + link

BDD/e2e suites require a live engine so they were not run here; the e2e harness is engine-config-driven and stays compatible (the engine delivers the inlined worker config as the `--config` seed).

## Notes

- The configuration dependency is implicit via runtime `iii.trigger` (no `iii.worker.yaml` deps entry), matching `database`.
- Pre-existing, unrelated `Cargo.lock` version bumps in other workers were intentionally left out of this branch.

## New Features
* Configuration is now managed by a dedicated configuration worker with hot-reload support for runtime settings (certain topology changes require restart)
* The `--config` flag is now optional; configuration can be provided via seed file or stored values
* Environment variable expansion is supported in configuration values

## Documentation
* Updated configuration documentation with details on hot-reload semantics, field behavior, and initialization options


iii-directory/src/functions/registry.rs
```diff
@@ -240,7 +241,9 @@ pub struct WorkerInfoOutput {
 #[derive(Clone)]
 pub struct RegistryCache {
     inner: Arc<RwLock<std::collections::HashMap<String, CacheEntry>>>,
-    ttl: Duration,
+    /// Live TTL in ms, shared with the worker-list cache and updated on a
+    /// `configuration:updated` reload (see `configuration::apply_config`).
+    ttl_ms: Arc<AtomicU64>,
```
guibeira, line 246: I'm not sure if we really need Arc in this case, since AtomicU64 implements sync and send, which means it's safe to share across threads. Do you mind trying to remove it and see if it compiles?

guibeira (approved):

iii-directory/src/functions/registry.rs
andersonleal: Good question — `AtomicU64` is indeed `Send + Sync`, but that only makes it safe to share a *reference* across threads; it doesn't give us *shared ownership*, which is what's needed here.

The same TTL cell is owned by three independent, `'static`-lived owners — `RegistryCache`, `RegisteredWorkersCache`, and `SharedState` — all built in `main.rs` from one `Arc::new(AtomicU64::new(...))`:

```rust
let cache_ttl_ms = Arc::new(AtomicU64::new(cfg.registry_cache_ttl_ms));
let registered_cache = make_registered_cache(cache_ttl_ms.clone());
let registry_cache   = RegistryCache::new_shared(cache_ttl_ms.clone());
// cache_ttl_ms then moved into SharedState
```

On a `configuration:updated` reload, `apply_config` does a single `cache_ttl_ms.store(...)`, and because all three point at the same allocation, both caches pick up the new `registry_cache_ttl_ms` immediately. A bare `AtomicU64` is a value, so each struct would hold its own copy and the store would only update one — hot-reload of the TTL would silently stop working.

It also wouldn't compile: `RegistryCache` (and `SharedState`) derive `Clone` and we actually clone them (`registry_cache.clone()`, `state.clone()`), but `AtomicU64` doesn't implement `Clone`:

```
error[E0277]: the trait bound `Atomic<u64>: Clone` is not satisfied
 #[derive(Clone)]
     ttl_ms: AtomicU64,
```

So the `Arc` is doing the shared-ownership job here rather than the thread-safety one, and it needs to stay. Thanks for the careful read!

PR #270, opened 2026-06-16
feat: publish typed JSON Schemas for llm-router + provider iii functions

ytallo:
## Why

The API reference at `workers.iii.dev/workers/<worker>?tab=api` renders **"unknown"** for `llm-router`, `provider-anthropic`, and `provider-openai`.

Root cause: every iii function in these workers is registered with a `Fn(Value) -> Value` handler. The SDK auto-extracts request/response schemas from the handler's argument/return types, but `serde_json::Value` produces the permissive **`AnyValue`** schema (no structure), so the registry has nothing concrete to render. (approval-gate renders fine because it uses typed `Fn(Req) -> Resp` handlers — this brings the router family to parity.)

## What

Attach **precise request/response JSON Schemas** at registration time via the SDK's `request_format`/`response_format`, **without changing any handler dispatch**. The handlers stay on `Value`, so every tolerant parse, streaming sink, bare-`null` answer, and error contract is byte-for-byte preserved.

- **llm-router**
  - Derive `JsonSchema` across the wire type graph (`events`/`messages`/`content`/`model`/`credential`/`router`).
  - New `wire_schema` module: `schema_of::<T>()` (mirrors the SDK's draft-07 generator) + `with_schemas::<Req, Resp>(reg)`.
  - Wrap all **14** `router::*` registrations. `router::models::get` publishes its real `{ model } | null` union; the engine-trigger handlers (`on_worker_available`, `on_config_changed`) publish a `null` response.
  - The precise request types already existed (`ChatRequest`, `ChatResponse`, `CompleteResponse`, …); added small schema-only types for the previously-inline shapes (`RouteRequest/Response`, `ModelsListRequest`, etc.) and the shared provider-protocol acks.
- **provider-anthropic / provider-openai**
  - Wrap the **3** `provider::*` registrations each, reusing `llm_router`'s helper and the shared `ProviderStreamInput` / `ProviderAck` / `RefreshModelsAck` / `NoParams` types. No direct `schemars` dep needed (the lockfile change is the transitive pickup via the path-dep).

`schemars` is pinned to `0.8` — the same major as `iii-sdk` 0.19.2 — so the emitted schemas are byte-identical to what a typed SDK handler would auto-extract, and there's a single `JsonSchema` trait across the graph.

## Behavior / risk

Schema-metadata only. `request_format`/`response_format` are descriptive registry metadata; the engine never validates payloads against them at invoke time, and the handler closures are untouched. An adversarial review confirmed dispatch is unchanged (handlers still take `Value`; no serde attribute changed).

## Tests

Golden `tests/schemas.rs` in each crate asserts every published request/response type renders a **structured** schema (never the `AnyValue` `true`), pins key fields (`ChatRequest` carries `writer_ref`; `CompleteRequest` does **not**), and checks the `{model}|null` and null-response shapes. All suites green: **llm-router 61, provider-anthropic 78, provider-openai 65** (incl. the existing engine-backed integration tests). `cargo fmt` + `clippy -D warnings` clean on all three.

## Release note for the operator

No `Cargo.toml` versions are bumped here (the `create-tag` workflow does that). Published schemas are collected by booting the worker at release time, and `release.yml` publishes **one worker per tag** — so after merge, run **Create Tag → bump `patch`** for **all three** workers (`llm-router`, then `provider-anthropic`, `provider-openai`) or the providers will keep serving their previously-registered (stale) schemas.

## Summary by CodeRabbit

* **New Features**
  * JSON schema publishing is now available for all API request and response types, enabling explicit documentation of wire contracts and improved tooling integration.

* **Tests**
  * Added comprehensive schema validation tests to ensure API contracts are properly documented and structured.

### DIFF
--- a/llm-router/src/types/content.rs
+++ b/llm-router/src/types/content.rs
@@ -1,7 +1,7 @@
 use serde::{Deserialize, Serialize};
 
 /// Content blocks — the atomic units of message content (README § Content blocks).
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, schemars::JsonSchema)]
 #[serde(tag = "type", rename_all = "snake_case")]
 pub enum ContentBlock {
     Text {
--- a/llm-router/Cargo.lock
+++ b/llm-router/Cargo.lock
@@ -774,13 +774,14 @@ checksum = "92daf443525c4cce67b150400bc2316076100ce0b3686209eb8cf3c31612e6f0"
 
 [[package]]
 name = "llm-router"
-version = "0.2.0"
+version = "0.2.1"
 dependencies = [
  "async-trait",
  "clap",
  "futures",
  "iii-sdk",
  "regex",
+ "schemars",
  "serde",
  "serde_json",
  "sha2",
--- a/llm-router/Cargo.toml
+++ b/llm-router/Cargo.toml
@@ -21,6 +21,9 @@ path = "src/lib.rs"
 iii-sdk = "=0.19.2"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
+# Same schemars major as iii-sdk 0.19.2 so the schemas we derive here are the
+# exact draft-07 shapes the SDK publishes for typed handlers (see wire_schema).
+schemars = "0.8"
 tokio = { version = "1", features = ["macros", "rt-multi-thread", "sync", "time", "signal"] }
 async-trait = "0.1"
 thiserror = "2"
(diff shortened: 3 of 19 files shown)


llm-router/src/types/content.rs
```diff
@@ -1,7 +1,7 @@
 use serde::{Deserialize, Serialize};
 
 /// Content blocks — the atomic units of message content (README § Content blocks).
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, schemars::JsonSchema)]
```
sergiofilhowz, line 4: can we remove the schemars:: and add it to the import?

sergiofilhowz (changes requested):

PR #297, opened 2026-06-19
fix(web): publish a typed response schema for web::fetch

andersonleal:
## Problem

The `web/v1.1.0` release ([run 27841712523](https://github.com/iii-hq/workers/actions/runs/27841712523/job/82402692354)) built fine but **failed at the registry publish step**:

```
untyped (AnyValue/empty) schemas in worker-interface.json: web::fetch.response_schema.
Register the handler with a typed struct deriving schemars::JsonSchema
```

`web::fetch` registers a raw `serde_json::Value` handler (deliberate: it owns the `invalid_payload` error contract and returns a polymorphic envelope). The SDK therefore emitted the permissive `AnyValue` schema for the **response**, which the publish gate (`collect_worker_interface.py --assert-typed-schemas`) rejects. The request schema was already typed via `.request_format(request_schema())`; only the response was untyped.

## Fix

- Add a `FetchResponse` struct deriving `JsonSchema` that documents the envelope (success / page-image / error variants; all fields optional since no single response carries all of them).
- Wire it with `.response_format(response_schema())` — symmetric with the existing `.request_format(request_schema())`, keeping the deliberate `Value`-handler design rather than rewriting the polymorphic envelope construction.
- Add a unit test asserting the response schema is a concrete object (`type`/`properties`), so this can't regress before CI.

`cargo test` (55), `clippy -D warnings`, and `cargo fmt --check` all pass.

## Follow-up after merge

`web/v1.1.0` was built from a commit without this fix, so re-running that release would fail again. After merging, cut a **new** web release (Create Tag → worker `web`, e.g. `v1.1.1`) to produce a passing publish.

https://claude.ai/code/session_01QY5uCFgwr4wXR7ZcAKanaR

### DIFF
--- a/web/src/functions/fetch.rs
+++ b/web/src/functions/fetch.rs
@@ -12,7 +12,7 @@ use serde_json::{json, Value};
 
 use crate::config::{SharedConfig, WebConfig};
 use crate::fetch::execute_fetch;
-use crate::schemas::{request_schema, FetchPayload, TOOL_DESCRIPTION};
+use crate::schemas::{request_schema, response_schema, FetchPayload, TOOL_DESCRIPTION};
 
 /// Parse + validate + execute. Always returns an envelope (never errors).
 pub async fn handle(payload: Value, cfg: &WebConfig) -> Value {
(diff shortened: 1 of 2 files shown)


andersonleal:
Superseded by #298 (same fix, rebased onto current main). Closing in favor of that PR.

PR #318, opened 2026-06-23
fix(pi): publish typed response schemas and strip the $schema meta-ref

rohitg00:
pi merged in #314 but cannot publish: the worker-interface CI gate and the registry validator both reject its schemas.

## What this fixes

1. **Untyped response schemas** — the interface collector rejects functions whose `response_schema` is the empty/AnyValue node. Every pi:: function declared only `request_format`, so `response_schema` was untyped. Declares a typed `response_format` on all eight functions plus `pi::on-config-change`.
2. **`$schema` meta-ref** — the registry publish validator has no draft-2020-12 meta-schema registered, so the `$schema` key `z.toJSONSchema` stamps at the root fails publish with `no schema with key or ref ...draft/2020-12/schema`. A `jsonSchema()` helper strips the root `$schema` from every request and response format, and `runtimeJsonSchema` does the same.

Status and sessions::list responses carry a concrete `SessionRecord` schema rather than an open record, so the published nested schema is fully typed.

## Validation

- Every emitted format verified against the collector's typed-keyword rule (passes) and confirmed to carry no `$schema` key.
- typecheck clean, lint clean, 69 tests pass.

## Note for the team

Both gates are new. claude-code and codex declare only `request_format` and do not strip `$schema`, so both will fail the same two gates on their next release. They are grandfathered until then.

### DIFF
--- a/pi/src/config.ts
+++ b/pi/src/config.ts
@@ -30,9 +30,12 @@ export type Config = z.infer<typeof ConfigSchema>;
 export const RuntimeConfigSchema = ConfigSchema.omit({ engine_url: true });
 export type RuntimeConfig = z.infer<typeof RuntimeConfigSchema>;
 
-/** JSON Schema published to the configuration worker. */
+/** JSON Schema published to the configuration worker. The registry validator
+ *  has no `$schema` meta-schema, so strip the draft-2020-12 `$schema` key. */
 export function runtimeJsonSchema(): Record<string, unknown> {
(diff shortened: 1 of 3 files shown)


rohitg00:
Superseded by a clean branch. #318 re-included the typed-schema commits already merged in #317 and carried a lint error from a rebase artifact. The only outstanding fix is the $schema strip, opened fresh off current main.

PR #333, opened 2026-06-25
feat(harness): agent trigger subscriptions — be notified instead of polling

ytallo:
## Why

Agents need a way to wait for "something happened" without polling in a loop. iii already has the right primitive for this: engine triggers. This PR lets an agent subscribe to an existing trigger type and receive a session notification when it fires.

The important constraint is that the caller that fires the trigger often does not know the agent session. The harness owns that mapping: when the agent registers a trigger, the harness binds the trusted `session_id` to the subscription and uses it later when the trigger fires.

## What

This adds agent trigger subscriptions through the existing engine API:

- `engine::register_trigger { trigger_type, config, label?, once? }`
  - intercepted by the harness
  - returns a harness `subscription_id` plus the effective `once` flag
  - registers the real engine trigger against the shared internal `harness::notify_agent` handler
  - stamps trusted registration metadata (`subscription_id`, `session_id`, `label`, `once`) that the engine stores on the `Trigger`
  - `once` defaults to true for one-shot-ish types (`state` / `stream` / custom trigger types) and false for recurring `cron`
- `engine::unregister_trigger { id }`
  - intercepted by the harness
  - treats `id` as the returned `subscription_id`
  - owner-checks the subscription before unregistering the underlying engine trigger
  - is idempotent for already-removed subscriptions

Example flow:

```text
agent -> engine::register_trigger {
  trigger_type: "state",
  config: { scope: "job", key: "42" },
  label: "job finished"
}

any producer -> state::set { scope: "job", key: "42", value: { done: true } }

engine -> harness::notify_agent
harness -> injects [notification: job finished] into the owning session
```

No harness-owned emit/bus is added. Emitting is whatever already produces the trigger, such as `state::set` for an ad-hoc state signal.

## How it works

- **Single invocation chokepoint.** Normal function calls route through `functions::subscribe::invoke`, so `engine::register_trigger` and `engine::unregister_trigger` can be intercepted consistently from the inline turn loop, `harness::function::trigger`, and deferred approval release paths.
- **Trusted session binding.** The model never provides `session_id`. The harness injects the owning session from the current turn/request context and stamps it into the trigger's registration metadata — the agent-supplied arguments can never widen the target function or the metadata.
- **Native metadata delivery.** The engine stores the registration metadata on the `Trigger` and delivers it to `harness::notify_agent` at fire time as a distinct invocation argument, separate from the fired payload — so the payload can never spoof the subscription id or session.
- **Shared fire handler.** Every subscription binds its engine trigger to the one internal `harness::notify_agent` function. On fire, the handler validates the delivered metadata, claims the fire against the local registry (liveness, ownership, and an idempotent notification entry id), tears down `once` subscriptions, and injects a `user`-role notification into the owning session — waking an idle session or steering a running turn, reusing the session's last turn options.
- **Lean local registry.** The engine owns the trigger and its fire-time metadata; the harness keeps only the bookkeeping it needs locally — the owning session (per-session cap, ownership checks, `session::deleted` cleanup), a fire counter for idempotent entry ids, and the engine-returned trigger id for teardown.
- **Ephemeral lifecycle.** Subscriptions are process-local and session-scoped. They are removed on explicit unregister, on first fire when `once` is true, on `session::deleted`, or by the engine's worker-disconnect GC on process exit. A per-session cap (64) limits active subscriptions.
- **Console support.** Notification entries carry a `{ notification: true }` origin so the chat UI renders them distinctly (bell indicator) instead of as ordinary user text.

## Test plan

- [x] `cargo test`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `pnpm typecheck`
- [x] focused Biome check for changed console files
- [x] `git diff --check`

## Notes / follow-ups

- Subscriptions are intentionally ephemeral for v1. If restart-surviving waits become necessary, persist subscriptions per session and re-register them on harness startup.
- The public surface intentionally reuses `engine::register_trigger` / `engine::unregister_trigger` instead of adding `harness::subscribe` / `harness::unsubscribe` wrappers.
- The harness forbids subscribing to `harness::*` trigger types so agents cannot bind harness internals or create self-notification loops.

## Summary by CodeRabbit

* **New Features**
  * Added trigger-based notifications so supported events can deliver messages into a session without polling.
  * Notification messages now appear distinctly in the chat UI with a bell indicator.
  * Session deletions now clean up related subscriptions automatically.

* **Bug Fixes**
  * Improved session transcript updates so message origins are preserved and notifications are correctly recognized.
  * Added safer handling for subscription limits and unsubscribe behavior.


harness/prompts/anthropic.txt
```diff
@@ -21,7 +21,11 @@ Consequences worth internalising:
   registration. Restarting a worker is invisible to callers as long as it re-registers the
   same function ids; two workers registering the same id load-balance automatically.
 - Triggers are the engine's push channel. NEVER poll (a timer re-reading a queue, file, or
-  table) when a trigger type fits — bind a trigger instead.
+  table) when a trigger type fits — bind a trigger instead. To be notified yourself, call
+  `harness::subscribe` with a trigger_type (cron, subscribe, state, stream, or a custom trigger type)
```
sergiofilhowz: i think we should remove these triggers, it can give ai a sense that `cron, subscribe, state, stream` are default ones, but they're like any other trigger type

```suggestion
  `harness::subscribe` with a trigger_type
```

PR #375, opened 2026-06-30
feat(shell): fold the selected session directory into the jail roots

ytallo:
## Problem

A console-selected working directory arrives at the shell worker as the
per-call `base_dir`, but the coder jail only permitted paths inside the
statically configured `host_roots`. Any selected directory outside them was
rejected with C215:

```
path is outside every allowed root: /Users/.../workers.
Allowed roots: /private/tmp.
```

So the working-directory picker could only ever scope a chat to a directory
that was already baked into the jail config — selecting a real project path
failed.

## Fix

`PathResolver::session_scoped` folds the selected `base_dir` into the
effective allowed roots **for that call** (via `with_session_root`), so every
downstream check treats it as a first-class root:

- containment in `resolve` / `resolve_in`
- the non-accessible denylist (which relativises through `containing_root`)
- `is_root` and `session_root`

It is a no-op when `base_dir` is absent, already inside a configured root
(only an `Arc` bump), or cannot be canonicalised (the handler's own
`resolve_in` then produces the precise error). Wired into every
`base_dir`-bearing coder dispatch closure (read / search / update / create /
delete / list / tree / move).

## Why this is safe to widen on

- `base_dir` is stamped by the harness control plane (workspace injection),
  **never by the model**, so only operator-chosen directories grow the jail.
- `resolve_in` still scopes access to the session directory.
- The non-accessible denylist still applies, because the selected directory
  is now a real root that `containing_root` can relativise against.

## Test plan

- [x] `cargo build` — clean, 0 warnings
- [x] `cargo test` — green; added `session_scoped_adds_selected_dir_outside_jail`
      covering: outside-jail dir rejected until added, reachable once added,
      `**/.env` still blocked under the added root, and no-op for an in-jail /
      absent `base_dir`.

## Follow-ups (out of scope)

- The selection is per-call and does not persist into `shell/config.yaml`.
- The console working-directory picker is still unconstrained by `host_roots`.

## Summary by CodeRabbit

* **New Features**
  * File and folder actions now respect the selected working directory for each request, making operations more consistent with the current session context.

* **Bug Fixes**
  * Improved path handling so valid session directories can be accessed even when they fall outside the default allowed roots.
  * Existing restrictions still apply, including blocked sensitive files and folders.

### DIFF
--- a/shell/src/code/functions/mod.rs
+++ b/shell/src/code/functions/mod.rs
@@ -292,6 +292,7 @@ fn register_read_file(iii: &IIIClient, resolver: Arc<PathResolver>, cfg: ConfigC
             let cfg = cfg.clone();
             async move {
                 let cfg = cfg.read().await.clone();
+                let resolver = resolver.session_scoped(req.base_dir.as_deref());
                 read_file::handle(resolver, cfg, req)
                     .await
                     .map_err(Error::from)
@@ -309,6 +310,7 @@ fn register_search(iii: &IIIClient, resolver: Arc<PathResolver>, cfg: ConfigCell
             let cfg = cfg.clone();
             async move {
                 let cfg = cfg.read().await.clone();
+                let resolver = resolver.session_scoped(req.base_dir.as_deref());
                 search::handle(resolver, cfg, req)
                     .await
                     .map_err(Error::from)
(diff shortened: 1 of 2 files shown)


ytallo:
Closing in favor of #376, which implements the workspace picker/control-plane solution independently.

PR #411, opened 2026-07-06
(MOT-3873) feat(console): code mode picker

ytallo:
### feat(console): code mode in the chat composer

Adds 'code' to the mode picker. A code-mode send omits the explicit
functions policy so the harness applies its curated native coding
toolset, and swaps in the coding identity prompt; the working-dir
fs_scope metadata is unchanged.

### fix(shell): coder::context resolves the fs_scope root, not the jail primary

The workspace snapshot ignored the harness-stamped filesystem scope and
always described the configured primary root, so a scoped session's git
state and instruction files were missing. The effective root is now the
canonicalized fs_scope root (session_root confinement) when present, and
instruction files resolve through resolve_opt against it.

### feat(harness): code prompt — environment block is pre-fetched, skip redundant coder::context

Fixes MOT-3873

### DIFF
--- a/console/web/src/lib/backend/harness-send.ts
+++ b/console/web/src/lib/backend/harness-send.ts
@@ -37,8 +37,12 @@ export type HarnessOutputContract =
   | { type: 'text' }
   | { type: 'json'; schema?: unknown }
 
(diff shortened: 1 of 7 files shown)


ytallo:
Consolidated into #423.

PR #412, opened 2026-07-06
(MOT-3874) feat(shell): coder::apply-patch + worktree add/remove

ytallo:
Adds the apply_patch (V4A) format codex-family models are trained on:
*** Begin/End Patch with Add/Delete/Update File hunks, @@ context
markers, optional Move to, and End of File pinning. seek_sequence and
the replacement engine are ported from OpenAI's codex-rs apply-patch
crate (Apache-2.0) so context matching (exact -> rstrip -> trim ->
unicode-normalised) matches the reference; the parser is a compact
re-implementation pinned by the reference test cases.

All-or-nothing: every hunk resolves through the jail and computes
before anything is written (context mismatch fails C210 with re-read
guidance, filesystem untouched); writes land atomically per file and
modified files return a bounded echo of the first changed region.

Refs MOT-3874


ytallo:
Consolidated into #423.

PR #414, opened 2026-07-06
(MOT-3874) feat(approval-gate): auto-mode trust for the read-only coder surface

ytallo:
Shipped default rules gain structured, auto-scoped allows for the
read-only coder functions (context / info / read-file / search / tree /
list-folder), so an auto-mode coding session prompts on writes and exec
instead of every read. Manual mode is unaffected — mode-scoped rules do
not match there — and writes (update-file, apply-patch, create/delete/
move, worktree-*) plus shell::exec still gate in every mode.

default_rule_values() is now the single source of truth for both the
compiled default specs and the configuration schema default, and the
rules schema items widen to oneOf string|structured so the shipped
structured defaults pass configuration::register validation.

Refs MOT-3874

### DIFF
--- a/approval-gate/src/config.rs
+++ b/approval-gate/src/config.rs
@@ -11,7 +11,7 @@ use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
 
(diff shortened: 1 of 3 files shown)


ytallo:
Consolidated into #423.

PR #415, opened 2026-07-06
(MOT-3874) feat(shell): post-write checks on coder writes

ytallo:
### feat(shell): post_write_checks — report-only diagnostics after coder writes

CoderConfig gains post_write_checks ({match_glob, command, timeout_ms},
default empty, hot-reloads like every coder knob): after a successful
coder::update-file / create-file / apply-patch write, each configured
check whose glob matches a written file's root-relative path runs ONCE
per call (deduplicated by command) via /bin/sh -c with the effective
root as cwd, and its bounded (4 KiB) output lands in the response's
top-level checks array. A failing, timing-out, or unrunnable check
never fails the edit — it is feedback for the caller to act on.
Config-sourced on purpose: workspace files must not choose commands.

Also promotes the fs_scope effective-root convention into
PathResolver::effective_root (context/worktree/checks now share it),
and both harness code prompts tell the model to read the checks array.

### fix(harness): worktree-isolated children are told to commit their work

The code prompt's no-commit rule left isolated children's worktrees
dirty, so clean-only removal kept them and the parent had nothing on
wt/<name> to merge. A worktree spawn now appends an explicit commit
instruction to the child's task — the sanctioned exception that makes
the branch the handoff artifact.

Fixes MOT-3874


ytallo:
Consolidated into #423.

PR #416, opened 2026-07-06
(MOT-3875) feat: routed provider family + freeform apply_patch custom tool

ytallo:
- harness: a send (or parentless spawn) that omits provider resolves it
  from the router's model catalog before options freeze, so prompt-family
  selection and routing agree — headless codex sends now get the
  apply_patch code identity without passing provider explicitly. Router
  unreachable → warn and fall back to the previous default-family
  behavior.
- provider-openai-codex: coder::apply-patch travels as a freeform custom
  tool named exactly apply_patch — the trained name and shape. The wire
  emits { type: custom }, and the SSE state machine accumulates
  custom_tool_call raw input (delta + authoritative done) into
  { patch: <raw text> }, so patches cross the wire with no JSON escaping.
  Harness and shell still see an ordinary coder::apply-patch call.

Refs MOT-3875

### DIFF
--- a/harness/src/functions/send.rs
+++ b/harness/src/functions/send.rs
@@ -121,7 +121,13 @@ pub async fn start(deps: &Deps, req: SendRequest) -> Result<StartOutcome, Harnes
     let session = deps.session().await;
 
     // Freeze the per-send options before moving the message out of `req`.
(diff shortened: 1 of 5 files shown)


ytallo:
Consolidated into #423.

PR #417, opened 2026-07-06
(MOT-3875) feat(shell): write journal + coder::undo / coder::checkpoints

ytallo:
Every mutating coder call (update-file, create-file, delete-file, move,
apply-patch) records bounded before-images into a per-root journal
(journal.dir / max_bytes 64MiB / max_records 500, hot-reloading,
oldest-first eviction, raw-byte blob sidecars — binary-safe). The
harness fs_scope stamp gains session_id/turn_id so records attribute to
the turn that wrote them.

coder::undo reverts the last N records or everything a turn changed:
restores run newest-first, every path re-validates through the LIVE
jail, files that did not pre-exist are removed, and the undo journals
its own pre-undo state first — undoing an undo redoes the change.
Unrecoverable gaps (oversized images, directory deletes/moves) are
reported in skipped, never silently ignored. coder::checkpoints lists
the journal (newest first) as the undo picker. Catalog grows to 15.

Refs MOT-3875

### DIFF
--- a/harness/src/deferred.rs
+++ b/harness/src/deferred.rs
@@ -134,6 +134,7 @@ pub async fn resolve(
                 arguments,
                 filesystem_root.as_deref(),
                 &trusted_roots,
+                Some((&record.session_id, &record.turn_id)),
             );
             if let Some(cp) = record.calls.get_mut(&req.function_call_id) {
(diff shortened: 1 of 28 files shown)


ytallo:
Consolidated into #423.

PR #418, opened 2026-07-06
(MOT-3875) feat(console): coder views for apply-patch, checks, context, worktrees

ytallo:
feat(console): coder views for apply-patch, checks, context, worktrees

Refs MOT-3875


ytallo:
Consolidated into #423.

PR #419, opened 2026-07-06
(MOT-3875) feat(harness): harness::todo session checklist

ytallo:
### feat(harness): harness::todo — the session's live task checklist

Full-list-replace todo semantics for multi-step coding runs: at most 50
items / 500 chars, exactly one in_progress, persisted to the
harness_todo state scope and mirrored into custom session entries
(type "todo", fresh entry id per update — session::append is a no-op
on repeated ids). In-turn calls get the CALLING session injected at the
invocation chokepoint, so a model can never write another session's
list. code_mode_policy gains harness::todo and both code prompts teach
the discipline; the console renders the latest todo entry as a live
checklist card (done-count header, status glyphs).

### fix(harness): provider resolution scans the router catalog

router::models::get requires a provider hint, so a provider-less send
resolved to None (default prompt family) — scan router::models::list
by model id instead.

### fix(harness): harness::todo accepts TodoWrite-shaped payloads

Models trained on TodoWrite-shaped todo tools send { todos: [{ content,
status }] }; serde aliases (todos->items, content->text) accept both
spellings, extra fields like activeForm are ignored as before.

Fixes MOT-3875

### DIFF
--- a/console/web/src/components/chat/Message.tsx
+++ b/console/web/src/components/chat/Message.tsx
@@ -89,6 +89,8 @@ export function Message({
     case 'system':
       return message.kind === 'compaction' ? (
         <CompactionMarker message={message} />
+      ) : message.kind === 'todo' ? (
+        <TodoCard message={message} />
       ) : (
         <SystemNotice message={message} />
       )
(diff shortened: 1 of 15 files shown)


ytallo:
Consolidated into #423.

PR #420, opened 2026-07-06
(MOT-3876) perf(harness): concurrent read-only dispatch + fewer model round trips

ytallo:
A live code-mode session spent ~30 of 66 minutes on standalone
harness::todo round trips (40 updates x ~45s of reasoning each) and
most call batches carried a single call. Function execution itself
measured milliseconds; the cost is round trips.

- Both code prompts now require todo updates to ride along with the
  next action calls (never a standalone message) and push the model to
  batch independent calls into one message.
- The dispatch loop runs in three phases: sequential policy/pre-hook/
  checkpoint pass, concurrent invoke for consecutive parallel-safe
  calls (read-only coder surface + bookkeeping via join_all), then
  in-order post-hook/append/done. Mutations and shell::exec stay
  strictly sequential. All Triggered checkpoints land in ONE durable
  write before any invoke, preserving at-most-once recovery.

Fixes MOT-3876

### DIFF
--- a/harness/Cargo.lock
+++ b/harness/Cargo.lock
@@ -375,12 +375,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "futures"
+version = "0.3.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


ytallo:
Consolidated into #423.

PR #421, opened 2026-07-06
(MOT-3877) feat(approval-gate): unclipped coder write payloads in approval excerpts

ytallo:
The 256-char excerpt clip made held coder writes unreviewable — a
clipped patch can't render as a diff in the approval card. Excerpts are
now function-aware: the five coder write functions keep their full
arguments (secret-key redaction still applies) under a 64 KiB budget
that falls back to classic clipping; everything else clips as before.

Refs MOT-3877


ytallo:
Consolidated into #423.

PR #422, opened 2026-07-06
(MOT-3877) feat(console): checkpoints dialog with per-turn undo + medium thinking default

ytallo:
Add a checkpoints dialog to the chat working-dir footer that lists the
shell's file-mutation journal newest-first, grouped per turn, and reverses
a group with one coder::undo call. Records authored by coder::undo are
labeled "revert" and offer "redo". The trigger is disabled until a working
directory is set. Also default the composer's thinking level to medium.

Fixes MOT-3877


ytallo:
Consolidated into #423.

PR #452, opened 2026-07-08
feat(harness,console): push queued-message events, drop the status poll (MOT-3837)

andersonleal:
## What

Removes the console's 2.5s `harness::status` poll for the queued-messages strip and replaces it with a push signal.

- **harness**: new `harness::message-queued` trigger type (third sibling of `harness::turn-started`/`turn-completed`, same binding config and fan-out), emitted after a message parks in the mid-turn `harness_queue`. Payload is a pointer (`session_id`, `entry_id`, `queued_at`), not the message.
- **console**: `turn-events-live.ts` gains `startQueuedEventsSubscription`; `ChatView` drops the `setInterval` and refetches `harness::status → queued` once per event (plus one catch-up fetch on stream start). Refetch-on-signal keeps the strip correct under at-least-once delivery.
- **docs**: new trigger type specced in `harness.md` § Trigger types emitted.

## Stacking note

`feat/message-queue` (MOT-3837, the queue itself) exists only locally, so this PR bases on `main` and **includes those commits** — my change is the last commit (`f7a9b35`). Land the queue work first, then this rebases to just the one commit.

## Related tickets

Because `feat/message-queue` rides along locally, this branch bundles the queue **and** a batch of unrelated console/harness work. Split into focused tickets for tracking:

- **MOT-3837** — queue messages + notifications while streaming; mid-stream queueing, editing, and in-place edit (the queue work).
- **MOT-3920** — triggers strip: collapse, inspect reaction spec, DAG flow view.
- **MOT-3921** — readable engine / function-call payloads in the console (double-encoded JSON, blank `functions::info` tab, rich `triggers::info`, `register_trigger` when→then).
- **MOT-3922** — harden reactive pipelines against silent fail-closed stalls (spawn-policy prompt guards across harness + 6 providers, unwritten-state-key annotation, react-task-vs-user-message fix).

## Checks

- `cargo test` (harness): 189 passed; `cargo fmt --check` + `clippy` clean
- console: `tsc -b --noEmit` clean; `vitest` turn-events-live 7/7 (2 new cases for the queued binding)
- biome: touched lines clean; 3 pre-existing findings in `ChatView.tsx` come from the base branch

## Not done (deliberate)

- The 5s triggers-list poll in `ChatView` stays — there's no register/unregister event to bind to today.
- No reconnect/keepalive fallback: a missed event only leaves the strip momentarily stale, and stream-end drains everything into the transcript anyway.


andersonleal:
Merged `origin/main` in (`d5dc67eb`) to pick up #448, which had landed on main and touched several of the same files as this branch.

Two real conflicts, both resolved by hand:
- **`Composer.tsx`**: #448 redesigned the toolbar (folded the thinking-level picker into `ModelPicker`, moved `AttachmentButton` after the model picker, switched to icon-only stop/send buttons). This branch's mid-stream-queue work (`queueWhileStreaming`, the `optionsDisabled`/`inputDisabled` split, the extra queue-send button) predates that redesign. Resolution: kept #448's new layout and combined `ModelPicker`, re-added the queue-send button in the new position, preserved the `optionsDisabled` locking semantics for the picker row.
- **`harness/src/functions/send.rs`**: #448 added a `message_preview` field threaded through `seed_or_merge`/`seed_new` (an OTel trace-tag improvement). This branch added the `try_enqueue` queue-path branching. Resolution: kept both — the queue path now also computes and passes `message_preview` on its non-queued (append) branch, matching #448's new signature.

Verified post-merge: `cargo build` + `cargo test` (190/190) clean, console `tsc`/`vitest` (856/856) clean, biome clean.

PR #464, opened 2026-07-09
(MOT-3944) feat(harness,queue): provision harness turn queue

ytallo:
Fixes MOT-3944

## Summary
- Adds durable named function queues to the standalone queue worker, including persisted definitions, retries, FIFO grouping, and DLQ handling.
- Defines harness-turn during harness startup before readiness.
- Routes all harness turn-loop work to that FIFO queue, grouped by session_id.
- Propagates trace context through queued delivery and removes legacy iii-queue dependencies and configuration entries.

## Behavior
- Harness startup fails if its queue cannot be ensured.
- Restored jobs wait for harness::turn to register before delivery.
- The queue uses concurrency 10, three retries, FIFO ordering per session, and the queue worker's default 30-minute invocation timeout.

## Validation
- cd queue && cargo test --lib — 122 passed.
- cd harness && cargo test — 200 passed across library, manifest, and schema suites.
- cargo fmt --check for both crates.
- Live codex/gpt-5.4 harness turn completed through harness-turn.
- Trace tree verified harness::send -> enqueue -> engine::queue::enqueue -> function_queue_job -> harness::turn.

Companion engine PR: https://github.com/iii-hq/iii/pull/1950

## Summary by CodeRabbit

* **New Features**
  * Added durable named function queues with FIFO grouped ordering, concurrency limits, priorities, retry/backoff, and dead-letter handling.
  * Added queue definition + enqueue capabilities (`queue::define` / `engine::queue::enqueue`).
  * Harness now provisions and uses a dedicated **harness-turn** queue with ensured startup for per-session ordering.
* **Bug Fixes**
  * Improved ack/nack, requeue, consumer stop/start behavior, and restart recovery to prevent lost/duplicated in-flight jobs.
* **Documentation**
  * Updated setup/operational guides and worker configuration examples to remove the old queue worker and reflect the new queue worker behavior.


queue/src/observability.rs
```diff
@@ -0,0 +1,52 @@
...
+//! span instead of capturing the startup no-op provider.
+
+use iii_helpers::observability::opentelemetry::global::{self, BoxedSpan};
+use iii_helpers::observability::opentelemetry::trace::{SpanBuilder, Tracer};
+use iii_helpers::observability::opentelemetry::Context;
+use tracing::Subscriber;
+use tracing_opentelemetry::OpenTelemetryLayer;
+use tracing_subscriber::registry::LookupSpan;
+
+const TRACER_NAME: &str = "iii-queue";
```
guibeira: let's switch to just queue

guibeira (approved):
Can we update only the harness and queue? i would like to first update those, and them we can run some tests. If everything works fine we can come back and update them.

PR #466, opened 2026-07-10
(MOT-3944) feat(queue): durable queues per function

ytallo:
Fixes MOT-3944

## What changes

Harness gives root conversations, sub-agents, and reactive turns their own durable function queues:

- `harness::turn::root`
- `harness::turn::subagent`
- `harness::turn::reactive`

The standalone `queue` worker owns provisioning, persistence, retries, and delivery. Harness sends turn work directly to that worker, so the legacy engine-owned `iii-queue` worker is not started.

## Why it matters

Busy sub-agent or reactive work no longer consumes the capacity reserved for new root conversations. Each lane has independent durable settings and consumer health reporting.

## Validation

- Queue unit and end-to-end coverage for function isolation, malformed messages, consumer recovery, and durability.
- Harness tests and workspace lint checks.
- Live stack verification: all three function queues were healthy with ten consumers each, accepted persisted work, and completed a root turn through the standalone queue worker.


ytallo:
Closing in favor of #464, which implements MOT-3944 with a single session-grouped `harness-turn` FIFO queue. The per-lane (root/subagent/reactive) isolation explored here can return as a follow-up if lane starvation proves to be a real problem in practice.

PR #513, opened 2026-07-16
(MOT-4053) feat(provider-llamacpp): native embeddings surface

rohitg00:
Adds `provider::llamacpp::embed`: batch text embeddings from the configured llama-server, which exposes an OpenAI-compatible `/v1/embeddings` when started with `--embeddings` and an embedding-capable model. The endpoint derives from the configured chat `api_url` (sibling path), the credential is optional to match the chat path (local servers usually run without `--api-key`), and the one-vector-per-input contract is enforced with count and index validation. Behind `router::embed`'s registry walk this gives the memory worker fully local semantic recall.

Live-verified on a rig: llama-server with a nomic-embed GGUF on a custom port; `provider::llamacpp::embed` directly and `router::embed { provider: llamacpp }` both returned correct 768-dim vectors, provider available in the registry. Wire schema golden committed; 67 tests green. The 3 integration tests that need an isolated engine fail identically on clean main on this host (environment) and are untouched.

Fixes MOT-4053

## Summary by CodeRabbit

- **New Features**
  - Added local batch text embeddings through the configured llama-server.
  - Supports one embedding vector per input while preserving input order.
  - Enables local semantic memory recall through the router without cloud calls.
  - Requires llama-server embeddings support and an embedding-capable GGUF model.

- **Documentation**
  - Documented embeddings setup, behavior, and requirements.

- **Security**
  - Restricted direct embedding calls so they can only be invoked through the router.

### DIFF
--- a/provider-llamacpp/Cargo.lock
+++ b/provider-llamacpp/Cargo.lock
@@ -738,7 +738,7 @@ checksum = "92daf443525c4cce67b150400bc2316076100ce0b3686209eb8cf3c31612e6f0"
 
 [[package]]
 name = "llm-router"
-version = "1.1.1"
+version = "1.1.3"
 dependencies = [
  "async-trait",
  "clap",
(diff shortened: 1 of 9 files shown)


rohitg00:
@coderabbitai review

PR #571, opened 2026-07-22
fix(worktree): stop failed lands leaking locked, prune-exempt worktrees

chrisayl:
## Problem

A land that hits `block()` — dirty target checkout, rebase conflict, test-gate failure, target moved past the retry bound, etc. — sets the record to `LandBlocked` but **never releases the worktree's creation-time git lock**. Only the `Finalize`/success path unlocks.

`prune` then **skips `LandBlocked` outright**, so a blocked worktree is simultaneously:
- permanently git-locked (nothing unlocks it outside a deletion path), and
- permanently exempt from the expiry sweep.

Net effect: every failed land leaks a locked worktree that accumulates forever and can only be cleared by hand.

## Fix

- **`block()` unlocks the worktree.** It stays on disk in `LandBlocked` for inspection/recovery, but no longer holds the git lock.
- **`prune` treats `LandBlocked` as eligible** for the same expiry sweep as `Active`/`Orphaned`. The existing clean + ahead/integration checks are unchanged, so a blocked worktree that still carries real unlanded work is preserved — only clean, integrated, expired ones are reaped.

## Test

Adds a `land_machine` regression asserting a blocked worktree is unlocked after `block()`. Full worktree worker suite green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Blocked land operations now release the associated worktree lock, preventing stale locked worktrees from impacting later inspection and cleanup.
  * Pruning now applies the standard reconciliation/cleanup rules to land-blocked records instead of skipping them unconditionally.
* **Tests**
  * Strengthened coverage to verify blocked worktrees are unlocked after W413 dirty-checked-out-primary blocks.
  * Added a prune scenario test ensuring blocked worktrees with unmerged work aren’t reaped until after integration.

### DIFF
--- a/worktree/src/functions/prune.rs
+++ b/worktree/src/functions/prune.rs
@@ -100,14 +100,10 @@ pub async fn handle(deps: &Deps, req: Request) -> Result<Response, WError> {
                 });
                 continue;
             }
-            Lifecycle::LandBlocked => {
-                skipped.push(PruneSkip {
-                    id: record.worktree_id,
-                    reason: "land-blocked".into(),
-                });
-                continue;
-            }
-            Lifecycle::Active | Lifecycle::Orphaned => {}
+            // LandBlocked joins the expiry sweep: block() unlocks the worktree,
+            // and the clean/ahead checks below preserve any that still hold
+            // unlanded work — only clean, integrated, expired ones are reaped.
+            Lifecycle::Active | Lifecycle::Orphaned | Lifecycle::LandBlocked => {}
         }
 
         let repo = Path::new(&record.repo_path);
--- a/worktree/src/functions/prune.rs
+++ b/worktree/src/functions/prune.rs
@@ -100,9 +100,6 @@ pub async fn handle(deps: &Deps, req: Request) -> Result<Response, WError> {
                 });
                 continue;
             }
-            // LandBlocked joins the expiry sweep: block() unlocks the worktree,
-            // and the clean/ahead checks below preserve any that still hold
-            // unlanded work — only clean, integrated, expired ones are reaped.
             Lifecycle::Active | Lifecycle::Orphaned | Lifecycle::LandBlocked => {}
         }
 
--- a/worktree/src/land/mod.rs
+++ b/worktree/src/land/mod.rs
@@ -460,10 +460,6 @@ async fn block(
     }
     state::clear_active_job_id(deps.state.as_ref(), &job.worktree_id).await?;
 
-    // Release the creation-time git lock so a blocked land does not leak a
-    // permanently locked worktree. The directory stays in place for inspection;
-    // the expiry prune sweep reaps it once clean, and manual removal no longer
-    // fights the lock.
     {
         let repo = Path::new(&job.repo_path);
         let wt = Path::new(&job.worktree_path);
--- a/worktree/src/land/mod.rs
+++ b/worktree/src/land/mod.rs
@@ -460,6 +460,18 @@ async fn block(
     }
     state::clear_active_job_id(deps.state.as_ref(), &job.worktree_id).await?;
 
+    // Release the creation-time git lock so a blocked land does not leak a
+    // permanently locked worktree. The directory stays in place for inspection;
+    // the expiry prune sweep reaps it once clean, and manual removal no longer
+    // fights the lock.
+    {
+        let repo = Path::new(&job.repo_path);
+        let wt = Path::new(&job.worktree_path);
+        if wt.exists() {
+            ops::worktree_unlock(repo, wt, deps.git_timeout_ms).await;
+        }
+    }
+
     tracing::info!(
         job_id = %job.job_id,
         worktree_id = %job.worktree_id,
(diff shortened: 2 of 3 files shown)


rohitg00 (changes requested):
Thanks for catching this. The leak is real: confirmed against current main, `block()` never unlocks and prune skips `LandBlocked` outright, so a blocked land leaves a worktree both git-locked and prune-exempt forever.

I verified the fix end to end on the branch: build clean, `cargo fmt --check` clean, `clippy -D warnings` zero, full suite green including the new unlock regression. I also exercised the prune eligibility change at runtime, since it is the part with data-loss potential and is currently untested: drove a worktree to `LandBlocked` with real unmerged commits, ran prune, confirmed it is preserved (`skipped: "unmerged work"`, still on disk), then integrated the work into the target and confirmed prune reaps it. The clean/ahead/integrated gates do preserve genuine unlanded work. Solid fix.

Three things before merge:

**1. Revert the version bump.** `Cargo.toml` goes `0.2.1 -> 0.2.4`, but main is `0.2.1` and `0.2.4` is already a published release tag. The release pipeline sets the version at release time, so this both fights that flow and collides with an existing tag (same version, different content). Leave it at `0.2.1`.

**2. Split or document the unrelated changes.** The description covers the lock/prune fix, but the diff also carries the `iii-state -> state` dependency rename, the clap `env` feature plus the `III_URL` arg, and the `iii.worker.yaml` `tags` field. The dependency rename is a legitimate migration and worth its own PR with that rationale; the env var and tags are fine but should be called out. Keeping the bugfix isolated makes it reviewable and safe to backport.

**3. Add a prune regression for the risky path.** The new test covers the unlock but not the prune eligibility change, which is the behavior with real consequences. This one locks in "preserve unmerged, reap integrated" so a future refactor cannot quietly turn the leak into deletion (verified passing against this branch):

```rust
#[tokio::test]
async fn prune_preserves_blocked_with_unmerged_then_reaps_integrated() {
    use worktree::functions::prune;
    let tmp = tempfile::tempdir().unwrap();
    let mut cfg = test_config(tmp.path());
    cfg.prune_expire_hours = 0; // everything is age-eligible
    let s = setup(make_env(tmp.path(), cfg), tmp.path()).await;
    let feature_sha = head_sha(&s.wt_path);

    // Reach LandBlocked via W413 (dirty primary): the worktree itself stays
    // clean, one commit ahead of main, and main is untouched (not integrated).
    std::fs::write(s.repo.join("README.md"), "local edits\n").unwrap();
    let job_id = enqueue_land(&s, "main", LandOverrides::default()).await;
    let deps = land_deps(&s.env).await;
    run_step(&deps, &job_id).await.unwrap();
    let rec = state::get_record(s.env.state.as_ref(), &s.worktree_id)
        .await
        .unwrap()
        .unwrap();
    assert_eq!(rec.lifecycle, Lifecycle::LandBlocked);

    // Unmerged work must be preserved, not reaped.
    let out = prune::handle(
        &s.env.deps,
        prune::Request { repo_path: None, dry_run: false },
    )
    .await
    .unwrap();
    assert!(!out.pruned.contains(&s.worktree_id), "must not prune unmerged: {out:?}");
    assert!(out
        .skipped
        .iter()
        .any(|k| k.id == s.worktree_id && k.reason == "unmerged work"));
    assert!(s.wt_path.is_dir());

    // Once the work is integrated, prune reaps it.
    let _ = git(&s.repo, &["merge", "--ff-only", &feature_sha]);
    let out2 = prune::handle(
        &s.env.deps,
        prune::Request { repo_path: None, dry_run: false },
    )
    .await
    .unwrap();
    assert!(out2.pruned.contains(&s.worktree_id), "must reap integrated: {out2:?}");
}
```

Revert the version and tighten the scope and this is a clean merge.

chrisayl:
Thanks for the thorough review. Addressed:

  3. Prune regression — added prune_preserves_blocked_with_unmerged_then_reaps_integrated in
  worktree/tests/land_machine.rs (1e1d734), matching what you outlined: unmerged work is preserved (skipped:  "unmerged work"), then reaped once integrated. Full suite green, cargo fmt --check clean, clippy --tests -D warnings clean.

  1 & 2 — checked this branch's diff against both its actual parent commit and current main (008917e): it touches only worktree/src/functions/prune.rs, worktree/src/land/mod.rs, and
  worktree/tests/land_machine.rs. No Cargo.toml version change (0.2.4 on both sides), no iii-state→state rename, no clap env/III_URL, no iii.worker.yaml tags field — none of that content exists anywhere in this branch's history. Not sure what ref that part of the review was diffed against, but there's nothing here to revert or split out. Ready to merge from my side.

rohitg00 (approved):
Verified the branch end to end: full `worktree` suite green, `land_machine` 13/13 including the new prune regression, `clippy --all-targets -D warnings` zero, `fmt --check` clean. The new test does what it claims, blocked with unmerged work is preserved as `skipped: "unmerged work"`, then reaped once the target absorbs it.

One follow-up worth its own PR, not a blocker here. `functions/land.rs` has no lifecycle guard, so a `LandBlocked` worktree can be re-landed, and on the `keep: true` path `Finalize` flips it back to `Active` without re-locking, since the lock is only ever taken at `worktree add --lock` in `git/ops.rs`. That leaves a managed `Active` worktree carrying no `iii:worktree` ownership lock. Nothing reads `lock_reason` today and prune reconciles stale records either way, so the impact is hygiene rather than data loss, but closing it needs a new `ops::worktree_lock`.

Approving.

PR #582, opened 2026-07-23
(MOT-4210) fix(harness): reject mismatched reaction specs across a join's predecessors

andersonleal:
## What

A join fires its downstream reaction with the **completing predecessor's** spec. Nothing stopped an agent from registering the full task on one predecessor and a shorthand on the rest — whichever arrived last silently replaced the real reaction.

Observed live (rctest-x7k2 postmortem): the orchestrator registered the full 7-step finalizer task on the `w1` join predecessor and literally `"task": "placeholder"` on `w2`/`w3`. `w3` completed the join, so the finalizer spawned with the seed message "placeholder" — its report, completion-signal, and cleanup instructions were all lost, and the run never communicated its end.

## How

The `JoinSpec` contract already required identical specs ("Every predecessor's subscription carries the same `id` + `expect` and its own `key`"); this enforces it at `engine::register_trigger` time, matching `validate_spec`'s reject-loudly posture:

- `react::join_canon` — fingerprints a join registration: metadata minus the per-predecessor `join.key` and harness stamps (`__subscription_id`, `__once`, `__owner_session_id`)
- `SubscriptionRegistry` — tracks the fingerprint per live entry (`JoinProbe`, attached post-insert like `set_trigger_id`; dies with the entry so retired predecessors never block re-registration) and surfaces the conflicting subscription id
- the subscribe interceptor rejects a divergent registration before binding, with an error naming the live twin and telling the agent to repeat the full spec

Join ids are globally namespaced (they key the shared accumulator record), so the conflict check scans across sessions. Fingerprinting runs post-`inherit_model`, so same-session registrations compare stamped metadata consistently. With enforcement in place, "completing predecessor's spec wins" becomes harmless — all predecessors are provably identical.

## Verification

- `cargo test` in harness: **264 unit + 7 integration pass, 0 failed**
- New tests pin the mechanism at both layers, including the exact rctest failure shape:
  - `join_canon_strips_per_predecessor_fields_only` / `join_canon_differs_when_the_reaction_differs`
  - `join_probe_conflict_detection_and_lifecycle`

Companion PR: #581 (state worker dropped the fire-time metadata sidecar — the other half of the rctest postmortem).

No version bump / changelog per repo convention.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to ensure all active predecessors for a join use the same reaction specification.
  * Registrations with conflicting reaction details are now rejected with an invalid request error.
  * Differences limited to predecessor-specific keys remain supported.
  * Conflict tracking is cleared when a predecessor subscription is retired.

### DIFF
--- a/harness/src/functions/react.rs
+++ b/harness/src/functions/react.rs
@@ -299,6 +299,33 @@ impl ReactResult {
     }
 }
 
+/// Canonical reaction-spec fingerprint for a join predecessor registration:
+/// the metadata minus this predecessor's own `join.key` and the harness
+/// stamps (which legitimately differ per predecessor). `None` when the spec
+/// is not a join predecessor.
+///
+/// All predecessors of one join id must agree on this fingerprint — the join
+/// fires with the COMPLETING predecessor's spec (`join_edge` → `spec.task`),
+/// so a divergent spec on a later predecessor (e.g. `task: "placeholder"`)
+/// silently replaces the real reaction. The registration interceptor rejects
+/// mismatches against live predecessors using it.
+pub fn join_canon(metadata: Option<&Value>) -> Option<(String, Value)> {
+    let m = metadata?;
+    let join_id = m.pointer("/join/id")?.as_str()?.to_string();
+    let mut canon = m.clone();
+    if let Some(obj) = canon.as_object_mut() {
+        // Stamps differ per registration by design; strip them defensively
+        // even though the interceptor fingerprints pre-stamp metadata.
+        obj.remove("__subscription_id");
+        obj.remove("__once");
+        obj.remove(crate::subscriptions::OWNER_SESSION_KEY);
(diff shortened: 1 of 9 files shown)


andersonleal:
Added **E2E-004 `join-spec-mismatch`** (`b83bc55f`): drives the exact postmortem shape through the real stack — full task on predecessor 1, `task: "placeholder"` on predecessor 2, same join id. Pins: first registration succeeds, second surfaces as an error `function_result` with the corrective text, model closes the turn on it. Verified meaningful: the scenario **fails against pre-fix sources**. All four direct scenarios pass locally (~3.2s each). Runner additions are generic: `allow_id`/`function_call_raw` (drive engine builtins through the scripted model), `messages_subset` (positional-subset history matching around nondeterministic content), transcript-level evidence helpers for intercepted builtins.

PR #586, opened 2026-07-24
(MOT-4217) fix(harness): let reaction sessions unregister their run's subscriptions

andersonleal:
## Why

rctest5-K7mQ ended in a cleanup deadlock. The repair reactor tried to tear down the run's subscriptions and hit `subscription belongs to a different session` ×3: `engine::unregister_trigger` is owner-session-scoped, and the owner — the orchestrator — was parked waiting on a `report` state notification that its own children could no longer satisfy. Result: 3 leaked armed subscriptions and a permanently sleeping session.

A reaction cleaning up its own run is the legitimate teardown path when the registrant is parked.

## What

- `spawn_reaction` records **child session → registrant session** lineage (from the trusted `__owner_session_id` stamp) in the ephemeral `SubscriptionRegistry`.
- The unregister ownership check accepts the owner itself **or any session whose lineage chain reaches the owner** (transitive, hop-bounded so re-targeted-session cycles terminate).
- Lineage entries are purged on `session::deleted` in both directions, keeping the map bounded by live sessions.

Deliberately narrow: strangers and the reverse direction (registrant "cleaning up" a child's subs it doesn't own) are still refused.

## Tests

- transitive grant (orchestrator → reactor → repair), reverse/stranger refusal
- self-edge ignored; A↔B cycle terminates without granting unrelated ownership
- `forget_lineage` purges entries pointing at and from the deleted session

267 harness tests pass; clippy clean.

Closes MOT-4217.

### DIFF
--- a/harness/src/functions/on_session_deleted.rs
+++ b/harness/src/functions/on_session_deleted.rs
@@ -24,6 +24,7 @@ pub async fn handle(
     event: SessionDeletedEvent,
 ) -> Result<SessionDeletedAck, HarnessError> {
     let dropped = deps.subscriptions.take_session(&event.session_id);
+    deps.subscriptions.forget_lineage(&event.session_id);
     let tracked = dropped.len() as u64;
     if tracked > 0 {
         for (_sub_id, trigger_id) in dropped {
--- a/harness/src/functions/react.rs
+++ b/harness/src/functions/react.rs
@@ -1246,6 +1246,13 @@ async fn spawn_reaction(
                 .get("child_turn_id")
                 .and_then(Value::as_str)
                 .map(str::to_string);
+            // Reaction sessions may need to clean up their own run while the
+            // registrant is parked on a wake: record child → registrant so
+            // the unregister ownership check accepts lineage descendants.
+            if let (Some(owner), Some(child)) = (spec.owner_session_id.as_deref(), child.as_deref())
+            {
+                deps.subscriptions.record_lineage(child, owner);
+            }
             tracing::info!(
                 child_session_id = child.as_deref(),
                 model = spec.model.as_deref(),
--- a/harness/src/functions/react.rs
+++ b/harness/src/functions/react.rs
@@ -546,10 +546,11 @@ fn stamp_coalesced(mut event: Value, dropped: usize) -> Value {
     event
 }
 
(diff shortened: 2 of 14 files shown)


andersonleal:
Restacked onto `feat/react-fire-coalescing` (PR #583) and extended with the e2e that gates this fix: **E2E-008 `reaction-unregisters-run`** — a standing binding's reaction, pinned to a separate session, unregisters the registrant's subscription; the router matcher demands the unregister result with `is_error: false` **and** `removed: true`. Reverting the ownership check to pre-lineage semantics times the scenario out; with the fix it passes (all 8 direct scenarios green). Also adds the router's serve-time capture primitive (`[[cap:…]]` — required because the `sub_…` id is runtime-generated) and an `armed_wake_advisory` on one-shot state-key wakes (the rctest5 orchestrator-parked-forever wiring gap).

andersonleal:
Third commit: **level-triggered join predecessors** (+ E2E-009 `late-join-predecessor-replay`).

Live rctest5 iteration exposed the registration-vs-completion race: a rejected divergent sibling (MOT-4210's check, working as designed) forced a re-registration round, the watched writer sessions finished inside that window, and the re-armed edge-triggered `turn-completed` join predecessors starved forever — finalizer never spawned, run parked.

- Registering a **join predecessor** on a `turn-completed` binding whose filtered session is already terminally completed (durable status `completed`, no armed wake) now delivers a **catch-up fire** shaped like the real completion event, stamped `__late_subscription_replay`. Joins only (per-key accumulator makes rare double-delivery idempotent); fail-open on status-lookup failure.
- **E2E-009** proves the completion is strictly in the past via a call-mode completion witness (recorder call #1), then a probe-steered turn registers the late join — the joined downstream's recorder call #2 exists only through the replay. Gate check: disabling the replay times the scenario out. Runner gains `ProbeAction.after_target_calls`.

**Live verification**: with this stack deployed (this PR + #583 + #585), the rctest5 orchestration prompt now completes fully green end to end — self-verifying pass report, totals == GROUP BY (5/5/5), 15/15 events, cited trigger-spawned reactor session, zero leaked subscriptions, autonomous cleanup. All 9 direct e2e scenarios pass.

PR #654, opened 2026-07-31
refactor(state): tenant-agnostic private namespaces replace hardcoded harness ids

andersonleal:
Stacked on `fix/harness-control-plane-review`: generalizes the private-scope facility that branch added to the state worker, removing every hardcoded `harness::state::*` id and `harness_binding*` scope from state code.

## What changed
- **`StateConfig.private_namespaces`** (restart-tier): workers claim namespaces — `{ functions_prefix, scopes }`. Each claim reserves its scopes (public `state::*` rejects them, trigger fan-out and `list_groups` skip them) and registers internal accessors `<prefix>::state::{get, list, compare-and-set}`.
- **Per-tenant hard scoping**: a namespace's accessors only reach its OWN scopes — tighter than the old any-reserved check (previously `harness::state::get` could read any reserved scope).
- **Loud boot validation**: empty/invalid prefixes, empty scopes, and duplicate scope claims are skipped with error logs; the first claimant keeps a contested scope.
- The harness's wire contract is untouched: with the config claim (`functions_prefix: harness`, scopes `harness_binding`, `harness_binding_owner`) the exact same ids exist — a parity test pins the generated ids; `harness/src/state.rs` needs no changes.

## Deployment note (breaking without config)
State deployments must add the claim to their state config or `harness::state::*` disappears (bindings fail loudly with `function_not_found`; the classic `fires: 0` symptom follows). Updated in-repo: `harness/tests/e2e/stack-config/state.yaml`. Dev stacks using the configuration worker's live value need the same entry added before restarting the state worker on this build.

## Tests
`state`: 128 tests green, including new coverage for reserved-scope guards with neutral tenant names, per-namespace access denial across tenants, id generation, and invalid/duplicate claim skipping.

### DIFF
--- a/harness/tests/e2e/stack-config/state.yaml
+++ b/harness/tests/e2e/stack-config/state.yaml
@@ -2,3 +2,9 @@ adapter:
   name: kv
   config:
     store_method: in_memory
+
+# The harness claims its binding-authority scopes; the state worker itself is
+# tenant-agnostic (StateConfig::private_namespaces).
+private_namespaces:
+  - functions_prefix: harness
+    scopes: [harness_binding, harness_binding_owner]
--- a/state/src/boot.rs
+++ b/state/src/boot.rs
@@ -51,6 +51,10 @@ pub async fn start(iii: Arc<IIIClient>, config: StateConfig) -> anyhow::Result<B
             .trigger_request_format::<StateTriggerSpec>(),
     );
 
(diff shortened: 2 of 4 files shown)


andersonleal:
Superseded: the state-worker generalization was cherry-picked onto `feat/state-control-plane` (MOT-4231, PR #653), which is the right home for the state-side work.

PR #686, opened 2026-08-03
(MOT-4327) feat(harness): live context accounting with a console chip

rohitg00:
## What

The harness accounts for every generation's context window and surfaces it live in the console.

Data plane:
- Each generate step folds the assembly it already performed into a `ContextSnapshotV1`: category token counts (system prompt, function schemas, messages by role, request overhead, post-assembly hook growth), the usable budget, the final request total, compaction state, and the estimator. Zero extra counting round trips; the categories come from context-manager's new assemble `breakdown` and degrade to totals-only against an older context-manager.
- After the terminal frame the snapshot is stamped with the generation's billed usage and stored under `harness_context/<session_id>` (one row per session, latest wins).
- When a provider token counter is available (`router::count_tokens`, #685), the snapshot becomes exact: billed usage is the total, the system prompt and tool schemas are counted through the provider's tokenizer with a probe delta and cached per model, provider and content, and the message window is the remainder. Rigs without a counter keep the heuristic numbers, and the `estimator` field always says which produced them.
- `harness::turn-completed` carries the snapshot as `context`; `harness::metrics.by_session[]` returns each session's latest snapshot, preferring the copy already on the turn record and falling back to the durable row for a freshly seeded turn.

Console surface (injectable UI shipped by this worker):
- A `context` session chip in the chat header: real percent of the usable window, live after every step. One `state::get` on mount, then streaming via the state worker's `state` trigger with an engine-side scope and key filter. No polling anywhere. Before the first turn it shows the model's catalog window.
- Click opens a breakdown popover: stacked category bar, legend, and a footer carrying provenance (`exact · provider tokenizer` or `est. heuristic`), fresh input and output, prompt cache read, write and hit rate, and cost. The hit rate is coloured by verdict, since a cold prefix re-bills the whole prompt at the write premium every turn.
- A function-trigger renderer for `harness::metrics` results in chat.

## Live verification (running rig)

Verified across three providers. Anthropic haiku: estimator `provider`, total 13,072 equal to billed to the token, system prompt 12,230 exact where the heuristic had said 8,262 (32 percent under). Codex gpt-5.4-mini: estimator `tiktoken`, total 11,481 equal to billed. Chip, popover, and mid-turn streaming verified in the console against live sessions.

The cache readout immediately earned itself: a three-turn probe showed a session pays two full cache writes before its first read, because the first generation's system prompt is 14 tokens larger than every turn after it. Filed as MOT-4347; not addressed here.

## Testing

316 tests including the integration scenario suite, wire-schema goldens regenerated, cargo fmt and clippy with `-D warnings` clean, UI bundle built and scope-asserted.

## Notes for review

- `crates/console-ui` relaxes its `iii-sdk` pin from an exact `=0.21.6` to a `0.21` range so harness can link it. That deleted a 314-line fork of the asset wire contract this PR originally carried, and each consumer still resolves its own exact pin in its own lockfile.
- `eval` follows harness onto `iii-sdk` 0.21.8. It path-depends on harness, so the two exact pins could never co-resolve once this branch moved; the eval CI lane fails on any harness-touching change without it.
- The integration probe's `LifecycleEventV1` is an exact wire shape and the scenario floor pins the payload key set, so both learn the optional `context` field. Without that, every scenario timed out waiting for a terminal status it had already received.
- Review cleanups from `/simplify` are in their own commits: `exactify` counts the probe base once and runs the two category deltas concurrently, `Assembled` carries `Applied` whole, and the chip drops a redundant `turn-completed` subscription that the state trigger already covers earlier and finer.
- Merge after #684 (chip slot; the chip feature-detects and skips silently on older consoles). Runtime exactness activates with #685; without it the snapshot stays heuristic by design.

Part of MOT-4324 (MOT-4327 and the harness half of MOT-4329).

## Summary by CodeRabbit

- **New Features**
  - Added a console UI showing live context usage, token counts, capacity, compaction details, cache activity, and cost.
  - Added detailed metrics displays with session-level usage, context bars, token formatting, and partial-status indicators.
  - Added provider-aware token counting and categorized context snapshots to improve usage reporting.
  - Context details are now included in completion events and session metrics when available.

- **Bug Fixes**
  - Improved compatibility with older or incomplete context data.
  - Automatically removes saved context information when a session is deleted.


harness/src/context_snapshot.rs
```diff
@@ -0,0 +1,402 @@
...
+    CACHE.get_or_init(|| Mutex::new(HashMap::new()))
+}
+
+fn cache_key(model: &str, kind: &str, content: &str) -> u64 {
+    let mut hasher = DefaultHasher::new();
+    model.hash(&mut hasher);
+    kind.hash(&mut hasher);
+    content.hash(&mut hasher);
+    hasher.finish()
+}
```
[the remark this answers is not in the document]
rohitg00: Fixed in 6a5988bf: cache_key now hashes the provider alongside model, kind, and content.

harness/src/functions/send.rs
```diff
@@ -687,6 +687,7 @@ pub(crate) async fn seed_new(
         reactive_depth: None,
         reactive_owner_session_id: None,
         functions_generation,
+        context_snapshot: None,
```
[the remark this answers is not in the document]
rohitg00: Fixed in 6a5988bf: metrics prefers the record snapshot and falls back to the durable harness_context row when a freshly seeded turn has not generated yet.

harness/src/turn_loop.rs
```diff
@@ -507,6 +507,14 @@ pub async fn run_step(
...
                 .unwrap_or(assembled.effective_max_output_tokens)
                 .min(assembled.effective_max_output_tokens);
+            let snapshot = build_context_snapshot(
+                &record,
+                payload.step,
+                &assembled,
+                final_request_tokens,
+                request_overhead_tokens,
+            );
+            record.context_snapshot = Some(snapshot);
```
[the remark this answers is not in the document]
rohitg00: Fixed in 6a5988bf: the unchanged path subtracts the unused reservation from the total.

harness/src/turn_loop.rs
```diff
@@ -2136,13 +2169,55 @@ async fn assemble_context(
...
+        effective_max_output_tokens: assembled.effective_max_output_tokens,
+        total: final_request_tokens,
+        free: assembled.usable.saturating_sub(final_request_tokens),
+        categories: SnapshotCategoriesV1 {
+            system_prompt: b.system_prompt_tokens,
+            tools: b.tools_tokens,
+            messages: b.by_role.into(),
+            overhead: request_overhead_tokens,
+            hook_guidance: final_request_tokens.saturating_sub(assembled.token_count),
+        },
```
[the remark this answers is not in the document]
rohitg00: Deferred: the assembly-time categories are estimates by design and the exactify path already corrects them whenever a provider counter exists. The clean fix is the tokenizer estimator inside context-manager (MOT-4329 follow-up), which removes the stale-category window for every consumer instead of special-casing hook rewrites here.

harness/ui/src/context-chip/index.tsx
```diff
@@ -0,0 +1,410 @@
...
+        .then((value) => {
+          if (cancelled) return
+          if (isSnapshot(value) && value.session_id === sessionId)
+            setSnapshot(value)
+        })
+        .catch(() => {})
+      return () => {
+        cancelled = true
+      }
+    }, [host, sessionId])
```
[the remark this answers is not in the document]
rohitg00: Fixed in 6a5988bf: both write paths go through an accept-newest guard keyed on the snapshot timestamp.

harness/ui/src/lib/metrics.ts
```diff
@@ -0,0 +1,117 @@
...
+  if (!value || typeof value !== 'object' || Array.isArray(value)) return false
+  const snap = value as Record<string, unknown>
+  return (
+    typeof snap.total === 'number' &&
+    typeof snap.usable === 'number' &&
+    !!snap.categories &&
+    typeof snap.categories === 'object' &&
+    !Array.isArray(snap.categories)
+  )
+}
```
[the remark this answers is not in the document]
rohitg00: Fixed in 6a5988bf: isSnapshot validates the messages object and its four numeric fields.

harness/src/context_snapshot.rs
```diff
@@ -0,0 +1,419 @@
...
+    timeout_ms: u64,
+) -> Result<Option<ContextSnapshotV1>, HarnessError> {
+    let v = crate::state::state_get(iii, CONTEXT_SCOPE, session_id, timeout_ms).await?;
+    if v.is_null() {
+        return Ok(None);
+    }
+    Ok(serde_json::from_value(v).ok())
+}
+
+pub async fn delete(
```
ytallo, line 169: `delete()` has no callers in this PR. `on_session_deleted::handle` purges filesystem grants and budget state, but not the new `harness_context/<session_id>` row, so deleted sessions leak their snapshot rows in the state worker. Should this be wired into `harness::on-session-deleted` alongside the other purges (or dropped, if cleanup is deliberately deferred to a follow-up)?

harness/src/context_snapshot.rs
```diff
@@ -0,0 +1,419 @@
...
+            }
+        },
+    );
+    let (Some(counted_system), Some(counted_tools)) = (counted_system, counted_tools) else {
+        return;
+    };
+    let estimator = counted_system
+        .estimator
+        .or(counted_tools.estimator)
+        .unwrap_or_else(|| "provider".into());
```
ytallo, line 353: Cache hits lose the estimator provenance: `counted_delta` returns `estimator: None` when served from cache, so once both parts are cached (every step after the first with a stable prompt and tools) this falls back to `"provider"`. On a tiktoken-backed rig — the Codex example in the PR description — step 2+ snapshots would read `exact · provider tokenizer` in the popover footer. Consider caching the estimator string alongside the token count (e.g. store `(u64, String)` in the cache).

harness/src/context_snapshot.rs
```diff
@@ -0,0 +1,419 @@
...
+            ..SnapshotMessagesV1::default()
+        }
+    };
+
+    snapshot.categories.system_prompt = counted_system.tokens;
+    snapshot.categories.tools = counted_tools.tokens;
+    snapshot.categories.messages = messages;
+    snapshot.categories.overhead = 0;
+    snapshot.categories.hook_guidance = 0;
+    snapshot.total = billed;
```
ytallo, line 378: `billed` can exceed `usable` (actual billed usage vs a budget derived from the heuristic estimate — e.g. the 32% system-prompt underestimate quoted in the PR description), at which point `total > usable` contradicts the documented invariant on `ContextSnapshotV1` ("`total <= usable` always holds for a generation that ran"), which also ships in the schema goldens' descriptions. `free` saturates and the UI clamps, so nothing breaks today, but schema consumers may rely on the stated invariant. Suggest softening the doc (or clamping `total` to `usable`).

harness/src/functions/metrics.rs
```diff
@@ -153,7 +159,18 @@ pub async fn handle(
...
             }
         }
-        by_session.push(current.finish_session(node));
+        // The turn record usually carries the session's latest snapshot for
+        // free; a freshly seeded turn has not generated yet, so fall back to
+        // the durable `harness_context` row (which the previous turn wrote).
+        let context = match turn.context_snapshot.clone() {
+            Some(snapshot) => Some(snapshot),
+            None => {
+                crate::context_snapshot::get(&deps.iii, &node.session_id, cfg.session_timeout_ms)
```
ytallo: This fallback runs one awaited `state::get` per session, serially, inside the tree walk. Any session whose latest turn record carries no snapshot pays the round trip — which today is every pre-existing session and every freshly seeded turn — so a large session tree adds sessions × RTT to each `harness::metrics` call, and partial snapshots are explicitly polled as progress signals (per the comment just below). Consider fetching these concurrently (`join_all`) or skipping the fallback on the polling/partial path.

harness/src/turn_loop.rs
```diff
@@ -720,6 +733,26 @@ pub async fn run_step(
     // advance()/finalize call that ends this step).
     record.watermark_entry_id = watermark;
 
+    // Stamp the generation's actual usage into the snapshot, replace the
+    // estimated categories with provider-exact counts where a counter
+    // exists, and store the session's latest copy. Best-effort: accounting
+    // must never fail a turn that generated successfully.
+    if let Some(snapshot) = record.context_snapshot.as_mut() {
+        snapshot.usage = outcome.message.usage.clone();
+        crate::context_snapshot::exactify(snapshot, &router, gen_system_prompt.as_deref(), &tools)
```
ytallo, line 742: `exactify` sits on the generation hot path: on a cache miss it is up to 3 counting round trips (probe base, then the two part deltas), each bounded only by the router client timeout, before the turn can finalize and emit `turn-completed`. With a stable prompt/tools that's first-step-only, but any per-step system-prompt change re-counts, and a slow or flaky counting endpoint delays every turn's completion. Worth a tighter dedicated timeout for `count_tokens`, or a comment on why the shared router timeout is acceptable here.

harness/ui/src/context-chip/index.tsx
```diff
@@ -0,0 +1,458 @@
...
+      const offHandler = host.iii.on<StateEvent>(STATE_FN, (event) => {
+        if (!event || event.key !== sessionId) return
+        if (event.event_type === 'state:deleted') return
+        if (isSnapshot(event.new_value) && event.new_value.session_id === sessionId)
+          acceptNewer(event.new_value)
+      })
+      const offTrigger = host.iii.registerTrigger({
+        type: 'state',
+        function_id: `${STATE_FN}::${host.iii.browserId}`,
+        config: { scope: 'harness_context', key: sessionId },
```
ytallo, line 364: The trigger's `function_id` is namespaced per tab (`browserId`) but not per session, and a handler/trigger pair is registered per mounted chip. If two chips are ever mounted at once (two sessions visible, or a mount/unmount race), both register the same `function_id` with different `key` configs — the second registration can clobber the first's engine-side filter, and one unmount's `offTrigger()` may tear down the other chip's stream. Including `sessionId` in the `function_id` would make the registrations independent.

harness/build.rs
```diff
@@ -1,6 +1,179 @@
...
+                    asset.display()
+                );
+            }
+        }
+        return;
+    }
+
+    let pnpm = locate_pnpm();
+
+    let status = Command::new(&pnpm)
```
ytallo, line 62: Heads-up: a plain `cargo build`/`cargo test` of the harness now hard-requires Node + pnpm (plus network for the first `pnpm install`), and the failure mode is a panic mid-build for someone touching only Rust. This matches the console worker's `web/` precedent per the module doc, so mainly: please confirm the Rust CI lanes and dev-setup docs already guarantee pnpm is present.

harness/src/context_snapshot.rs
rohitg00: Good catch, it was dead code. Wired into on-session-deleted in 085fac77, right next to the grants and budget purges.

harness/src/context_snapshot.rs
rohitg00: Yep. The cache holds (tokens, estimator) now, so a hit keeps the provenance. Probe base stores None deliberately, its estimator never reaches a snapshot anyway.

harness/src/context_snapshot.rs
rohitg00: Right, billed can exceed usable once usage lands. Went with softening the doc rather than clamping, since clamping would hide the overage and that is the one number worth seeing. Golden regenerated.

harness/src/functions/metrics.rs
rohitg00: Fair. Took your second option: the indices are collected during the walk and read once for the final complete response, so polled partials pay nothing.

harness/src/turn_loop.rs
rohitg00: fixed

harness/ui/src/context-chip/index.tsx
rohitg00: Fixed, the id carries the session now. One wrinkle worth noting: browserId has to stay the last segment, since on() namespaces the handler with it internally, so it ended up as STATE_FN::sessionId::browserId.

harness/build.rs
rohitg00: Checked, both lanes are already covered. ci.yml and release.yml key on <worker>/ui/package.json, and _rust-binary.yml pre-builds the bundle on Linux so the cross-compile shards never need a node toolchain. Same wiring as console, editor, database and pdf. The real gap is that nothing tells a rust-only contributor a harness build now needs pnpm, and that is equally true of the other four, so I left it out of this PR.

PR #687, opened 2026-08-03
(MOT-4335) feat(harness): parent-owned control plane and the post-turn validation hook

andersonleal:
## What this does

Makes the **parent agent the only owner of the control plane**, and adds a **post-turn validation hook** so a turn's completion can be gated by a function instead of by prose in a prompt.

A trigger binding now has exactly two shapes: **wake the owner** (omit `function_id` — the event arrives as a message and starts a turn) or **call a plain function** (deterministic, token-free, result discarded). It can never start an agent. Every sub-agent is started by a live parent through `harness::spawn`, so delegation always has an owner, a policy ceiling, and a place for failures to surface.

`harness::hook::post-turn` is the sixth hook point: it runs synchronously inside finalize after the output contract accepts a result, and a deny re-prompts the turn (bounded by `max_validation_retries`). It attaches through `engine::register_trigger`, and it is the one harness-internal trigger type an agent may bind for itself — with `sessions` force-stamped to its own session and its own children, so a validator can never gate a stranger's turns.

## The five follow-up fixes

Each came from watching real runs on a live stack today, in order:

| commit | what it fixes |
|---|---|
| `55d8740a` | provider identity prompts drifted from the harness default; resynced and gated off by default |
| `92fbeb1f` | agent-chosen spawn ids repeat across runs (models are not RNGs), and silent reuse hijacked another run's session — retained its transcript and left the console nesting under the original parent. In-turn reuse is now confined to the caller's own tree, and `reused` is reported |
| `4510ac0f` | a binding whose condition call *errors* skipped every fire while the diagnosis landed only in a transcript the parked owner never reads — three runs slept forever. The owner is now notified once per binding |
| `71c0b215` | the sub-agent contract mandates an `engine::functions::info` round, but parents narrow children to just the work functions — so the *obedient* child was policy-denied its mandatory first step while siblings that skipped discovery succeeded. Every child now keeps the discovery pair |
| `64e91ded` | prompt doctrine for the judgment slips the harness can't catch: arm the watch before starting its producer, reconcile once after any re-registration, always deadline a wake the run depends on, and keep task-named table/scope/key byte-identical to what the bindings watch |

## Compatibility

- **Bindings can no longer target `harness::spawn`.** Records that predate this are retired on both sides and their owner is notified; a fire already in flight is caught by the same backstop.
- **Turn events are not bindable** (`harness::turn-started` / `harness::turn-completed`): a session notified of its own turn ending would wake itself forever. Watch what the work *writes* instead.
- **A spawned child is a leaf by default** — its policy gains deny globs for `harness::spawn`, `harness::send` and trigger registration. Pass `options: { orchestrator: true }` for a child that must coordinate further agents.
- **Reactive subscriptions must be re-registered** after deploy.
- An in-turn spawn naming an **existing** session outside the caller's own tree is now refused (with the owner named).
- Every spawned child's dispatch policy now also allows `engine::functions::list` / `::info`.

## Testing

Rebased onto `28fab498` and re-verified end to end:

- **Unit / contract:** 300 harness lib, 85 integration-crate, 84 e2e-crate, 5 prompt lints, 3 wire-schema goldens. `clippy -D warnings` clean across all three workspaces.
- **Deterministic integration suite** (`harness/tests/integration`, no provider key): **16/16 direct scenarios pass**, including three new fixtures written for the fixes above — `INT-018 spawn-reuse-guard`, `INT-019 condition-failure-notice`, `INT-020 child-discovery-granted`. Each was proven to **fail on the unfixed harness** before being accepted.
- **Judge-backed e2e suite** (`harness/tests/e2e`, 19 scenarios, one isolated stack per scenario) run locally with **glm-5.2 as both subject and judge** — CI's default subject is a stronger model, so local scores are glm-calibrated and the pass/fail line sits lower than CI's. Full table posted as a comment; every scenario scored at or above its threshold except one noted below.

Fixes MOT-4335

### INT-005 — found and fixed

`INT-005 direct-spawn-leaf-pipeline` was failing in CI (collect-phase hang) and locally (generation-2 match failure). One cause behind both: the fixture armed a `state` wake gated by a `state::barrier` condition but never granted itself `state::barrier`, so `engine::register_trigger` refused the binding — *"a reaction can only call functions you can call yourself"*. Nothing was armed, the parked root was never woken, and the trace oracle waited forever for a second turn id that could not exist. The harness rule is correct; the fixture was under-declared. Fixed by granting the barrier, and the deterministic suite is now **16/16 green**.

Worth recording that an earlier claim of mine in this PR was wrong: I described INT-005 as a pre-existing local artifact on the strength of an A/B that stashed only the later commits, never the commit that introduced the fixture. It was mine the whole time.

### Known failures not introduced here

- `provider-llamacpp`'s `provider_registers_with_persisted_token_and_discovers_catalog_without_a_credential` fails in this local environment; **main's own copy fails identically**, so it is environmental and unrelated.
- `INT-010 crash-recovery-507` flakes under full-sweep CPU load and passes standalone (7s). It SIGKILLs and restarts the engine, so it is load-sensitive by nature.
- `receiving_operation` (e2e) scores 45/90 locally: glm-5.2 renames the courier sessions the prompt names exactly (`…-acme` → `…-acme-k3m7`), so the evaluator cannot find the couriers where it was told they would be. The identity prompt has been corrected to require a given id be passed verbatim, and the rebuilt binary carries that text, but this subject still ignores it — model non-compliance, not plumbing. The gates are right and stay as they are: obeying an explicitly given id IS the requirement.
- `validation_self_repair` (e2e) failed one local run as `subject_error`: glm-5.2 spent all 14 turns without repairing the seeded dataset, so the post-turn gate refused the turn at the cap. That is the intended behavior added here — the `max_turns` guard runs the post-turn chain instead of finalizing `completed` past it — and it is the one scenario deliberately built so the validator only *diagnoses* and the model must invent the fix. It scored 100 on an earlier sweep of the same suite with the same model, so this is subject-model variance, not a harness regression: 17 function calls, zero function-call errors, 1.1k output tokens across 14 turns (a weak-model spin, not a stuck harness).

## Review note

This is a large diff (196 files) because the control-plane change touches the harness, its prompts, the console renderer, and both test suites at once. The six commits are individually coherent and worth reading in order — the feature first, then each fix with its regression fixture.

### Console Playwright — root-caused and fixed

The four failing Console specs were pre-existing tests this branch never touched; the regression was mine anyway, one layer down: the branch added the **database worker** to the required integration stack, the playground driver shares that stack, and four enumerations of the playground worker set were never updated. The runner exited at boot (`runner_error: missing --worker-bin for: database`), no ready manifest appeared, and every spec timed out identically waiting for it. Fixed in the Playwright fixture, the CI workflow env, the make target, and the README (`a41c9e17`); the full Console suite passes locally in 15.5s, 4/4.


andersonleal:
## Judge-backed e2e suite — full local sweep

19 scenarios, one isolated stack each, **glm-5.2 as both subject and judge** (the only key on this box; CI's default subject is stronger, so these scores are glm-calibrated and the bar sits lower than CI's). **17 pass / 2 fail**, ~$5.95.

| scenario | verdict | score / threshold | time |
|---|---|---|---|
| direct_answer | pass | 100 / 50 | 5s |
| design_tradeoff | pass | 100 / 50 | 27s |
| security_triage | pass | 79 / 50 | 16s |
| security_review | pass | 100 / 50 | 12s |
| persistent_state | pass | 100 / 90 | 9s |
| timer_wake | pass | 100 / 90 | 25s |
| mechanical_reaction | pass | 100 / 90 | 29s |
| validation_loop | pass | 100 / 90 | 18s |
| custom_validator | pass | 100 / 90 | 13s |
| validation_chain | pass | 100 / 90 | 53s |
| **validation_self_repair** | **fail** | n/a / 80 | 45s |
| validation_scope_enforcement | pass | 100 / 90 | 17s |
| subagent_validation | pass | 100 / 90 | 41s |
| subagent_validation_failure | pass | 100 / 90 | 3m14s |
| multi_subagent_validation | pass | 100 / 90 | 1m16s |
| research_pipeline | pass | 100 / 90 | 1m42s |
| shell_coder_sandbox | pass | 55 / 50 | 1m32s |
| **receiving_operation** | **fail** | 45 / 90 | 2m43s |
| reactive_automation | pass | 100 / 90 | 1m57s |

The two new scenarios this branch rebased onto (`design_tradeoff`, `security_triage`) pass on their first run against it.

### The two failures are subject-model behavior, not harness defects

**`validation_self_repair` — `subject_error`.** glm-5.2 spent all 14 turns and 17 function calls (zero call errors, 1.1k output tokens total: spinning, not stuck) without repairing the seeded dataset, so the post-turn gate refused the turn at the cap. That refusal is a feature this PR adds — the `max_turns` guard now runs the post-turn chain instead of finalizing `completed` past it. This is the one scenario built so the validator only *diagnoses* and the model must invent the fix; it scored 100 on an earlier sweep with the same model.

**`receiving_operation` — 45/90, three gates red** (`minimal_spawns`, `courier_execution`, `no_late_root_repair`), all from one cause: the model renamed the courier sessions the prompt names exactly (`…-acme` → `…-acme-k3m7`), so the evaluator cannot find the couriers where it was told they would be. The identity prompt now requires a given id be passed verbatim (`d290342d`), and the rebuilt binary carries that text, but this subject still ignores it. The gates stay as they are — obeying an explicitly given id *is* the requirement, and loosening them would encode style over contract.

Amusing corroboration of the spawn-reuse guard this PR adds: glm-5.2 reinvented the identical "random" suffix `k3m7` for the acme courier across four separate runs today. Models are not RNGs, which is exactly why silent session reuse needed a mechanical wall rather than a prompt warning.

PR #696, opened 2026-08-04
chore: regenerate provider lock files for llm-router 1.4.1

ytallo:
## Summary
Sync all nine provider workers' `Cargo.lock` files to the llm-router **1.4.1** bump (531b1407), which updated only the router's own manifest. Every provider pins llm-router as a path dependency, so their stale locks (1.4.0; `llamacpp`/`xai` still at **1.3.2**) break any `--locked` build.

## Impact
First Harness E2E run on main after the bump ([30955188907](https://github.com/iii-hq/workers/actions/runs/30955188907)) failed in the **build** job before any scenario started:

```
error: cannot update the lock file .../provider-deepseek/Cargo.lock because --locked was passed
```

Every provider in the E2E matrix (subject `deepseek`, judge `zai`) is affected — deepseek just failed first. All nine locks now resolve cleanly under `--locked` (`cargo tree --locked`).

## Follow-up worth considering
The `llamacpp`/`xai` locks were stale across *two* bumps — the worker-bump tooling never regenerates dependent providers' locks. A bump-time hook (or CI check) that runs `cargo update -p llm-router` across `provider-*/` would prevent recurrence.

### DIFF
--- a/provider-anthropic/Cargo.lock
+++ b/provider-anthropic/Cargo.lock
@@ -668,6 +668,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "iii-console-ui"
+version = "0.1.0"
+dependencies = [
+ "iii-sdk",
+ "schemars",
+ "serde",
+ "serde_json",
+ "tokio",
+ "tracing",
+]
+
+[[package]]
 name = "iii-helpers"
 version = "0.21.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


ytallo:
Superseded by the feature-gate PR — the lock regeneration alone exposed that providers now need a Node toolchain to compile llm-router 1.4.1's build.rs; the successor PR fixes that structurally and carries these same lock updates.

PR #742, opened 2026-08-07
(MOT-4377) feat(workers-dev): named stacks — select a default, create and manage them from the TUI

andersonleal:
Closes MOT-4377.

`workers-dev` had one hardcoded "harness stack", and defining a different one meant quitting the TUI to hand-edit YAML. This adds named stacks you can select, and then create and manage without leaving the dashboard.

## Selecting a stack

- `<repo_root>/workers-dev.yaml` now **auto-loads** (no `--config` needed); `--config` still overrides it.
- New keys: `stacks:` — a mapping of `<name>: [roots…]`, YAML order preserved — and `default_stack:`. A stack's values are *roots*, expanded to roots + transitive deps for both grouping and starting, so a newly declared dependency joins automatically.
- A built-in `harness` stack always exists; a `stacks.harness:` entry overrides its roots.
- The dashboard groups by a session-scoped **current stack** (`── stack:<name> (N) ──`, full width). `Ctrl+u` starts the default when only one stack exists, and opens a picker when there are several — Enter switches the grouping **and** starts.

## Creating and managing stacks

- `Space` marks the selected worker (marked rows show `✓`; the table title counts them). `n` names the marked set as a new stack.
- Saving makes it the current stack immediately, with no restart — and deliberately does **not** start it. Press `Ctrl+u` when you want that.
- In the picker: `x` deletes a stack (confirm dialog naming the file), `*` makes it the default. Deleting the default is refused; deleting the current stack falls back to the default.

## How writes work

Every write is: **edit the YAML text in place → validate by re-parsing with the real loader → write a temp file beside the target → `rename` over it.** Comments, blank lines and key order survive because nothing is re-serialized. A failed write changes nothing — not the file, not the session state.

The scanner was aligned to YAML's own rules rather than approximations of them: `s-white` is space and tab only, `#` starts a comment only at line start or after whitespace, and `-` is a sequence indicator only before whitespace. That came out of fuzzing 40,000 generated documents against `serde_yaml` with a zero-byte-file invariant asserted on every success — unsafe outcomes went 768 → 0.

Deliberate refusals (refusing beats mangling, and each says so): inline `stacks: {…}`, duplicated top-level keys, multi-line flow sequences, nested mapping values, `? key`, tab-indented keys, `tiny :`, `-weird:`, and quoted keys on the delete path.

## Breaking change

`harness_stack:` is **hard-removed**. A local `workers-dev.yaml` still using it fails startup with a rename hint pointing at `stacks: {harness: [...]}`. Nothing in the repo commits such a file — only per-worktree ones are affected.

## Testing

109 tests in `workers-dev`; `cargo fmt --check`, `cargo clippy --all-features -- -D warnings` and `--all-features --all-targets -- -D warnings` all clean.

Manual QA was run live under tmux against a real engine: creating a stack from marks and confirming every comment in the file survived, `*` moving the default (including preserving a trailing comment on that line), the delete guardrail refusing the default, deleting a non-default stack, and the inline-`stacks:` refusal leaving the file untouched.

**Not exercised:** pressing Enter in the picker (switch *and* start) against a live engine — it spawns real `cargo run` builds for every root, against an engine another session owned. The path is unit-tested and traced end to end, but not observed.

## Notes

- Purely additive: with no config file present, and for anyone who never presses `Space`/`n`/`x`/`*`, behavior matches the previous release apart from deliberately reworded status strings (`started harness stack` → `started stack harness`).
- `workers-dev.yaml` is not gitignored, so the first save leaves an untracked file in the repo root; the save notice names the path it wrote and the README says so.
- No version bump and no CHANGELOG, per the workers-ci bot's ownership of versions.

## Summary by CodeRabbit

- **New Features**
  - Added named worker stacks with configurable roots and transitive dependency discovery.
  - Added stack selection, creation, deletion, default-stack management, and startup controls in the TUI.
  - Added worker marking and stack-saving interactions.
  - Status views and dashboards now display active stack membership and labels.

- **Configuration**
  - Replaced the harness stack setting with configurable stacks and a default stack.
  - Added automatic configuration-file discovery and safer configuration editing.

- **Documentation**
  - Updated CLI, TUI, configuration, and startup behavior documentation.

### DIFF
--- a/workers-dev/README.md
+++ b/workers-dev/README.md
@@ -127,6 +127,7 @@ workers:          # optional override; default = all discovered
   - session-manager
   - llm-router
   - harness
+  - console
 stacks:           # optional named stacks; values are stack ROOTS —
   console:        # the group shown/started is roots + transitive deps.
     - console
--- a/workers-dev/README.md
+++ b/workers-dev/README.md
@@ -30,7 +30,7 @@ Workers are **discovered automatically** from top-level `*/iii.worker.yaml` in t
 
 | Group | Workers | Started by |
 |-------|---------|------------|
-| **harness stack** | The stack roots (`session-manager`, `llm-router`, `context-manager`, `provider-anthropic`, `provider-openai`, `approval-gate`, `harness`) **plus everything they transitively depend on**, derived live from each worker's `iii.worker.yaml` dependencies | `workers-dev up`, `Ctrl+u` in TUI, `workers-dev start` (starts the roots; missing deps are pulled in, connected ones left alone) |
+| **current stack** (default: `harness`) | The stack's roots **plus everything they transitively depend on**, derived live from each worker's `iii.worker.yaml` dependencies. The built-in `harness` stack's roots are `session-manager`, `llm-router`, `context-manager`, `provider-anthropic`, `provider-openai`, `approval-gate`, `harness`; define more stacks (or override `harness`'s roots) in `workers-dev.yaml` — see Config below | `workers-dev up` / bare `workers-dev start` start the `default_stack` (missing deps are pulled in, connected ones left alone); `Ctrl+u` in the TUI starts it directly, or opens a stack picker when more than one stack is defined |
 | **other** | All remaining repo workers (e.g. `telegram-bot`, `console`, …) | `workers-dev start <name>`, `workers-dev start --all`, `Ctrl+a` in TUI |
 
 Press `d` on any worker in the TUI to see its direct dependencies and its transitive dependents (the `r` restart blast radius), each with live status.
@@ -42,9 +42,9 @@ A worker connected to the engine but not started by this `workers-dev` shows **P
 ## Usage
 
 ```bash
-workers-dev up                    # start harness stack + TUI
+workers-dev up                    # start default stack + TUI
 workers-dev                       # TUI only
-workers-dev start                 # harness stack (CLI, waits for connect)
+workers-dev start                 # default stack (CLI, waits for connect)
 workers-dev start --all           # every discovered Rust worker
 workers-dev start telegram-bot    # one worker (+ missing deps)
 workers-dev restart llm-router    # rebuild + restart dependents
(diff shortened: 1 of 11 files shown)


andersonleal:
Thanks — all four actionable comments and all three nitpicks were valid and are addressed in `496be4d0`..`2220b9d0`.

**Reject an empty default stack before starting the engine** (`commands/mod.rs`) — fixed in `ed4ba6ee`. `run_up` now resolves and validates the default stack's roots before `ensure_engine()`, so no `iii` process is spawned for a config that cannot start anything. The `start_roots` guard is untouched, since the TUI path depends on it.

**Reject terminal control characters in configured stack names** (`status.rs`) — fixed in `b39995a2`, and this was the best catch of the review. Auto-loading `<repo_root>/workers-dev.yaml` is new on this branch, so cloning a repo and running `workers-dev status` renders whatever that file's keys contain, and a double-quoted YAML key can carry `\e`. Validated at load rather than at render, so every consumer is covered by one rule: `parse_stacks` and `default_stack` now reject names containing control characters, naming the offending key.

**Use a unique temp file and don't claim crash durability from the rename** (`config_write.rs`) — fixed in `45d31d61`. The temp path now carries the pid, so concurrent instances writing the same config can't clobber each other or delete one another's temp file on an error path. Rather than weaken the doc comment, the claim was made true: the write goes through a `File` with an explicit `flush()` + `sync_all()` before the rename.

**Preserve orphaned block comments when the last entry is deleted** (`config_write.rs`) — fixed in `496be4d0`. The `stacks:` header is now kept when the block still holds comment lines. Deliberately not the other option on offer: this module's contract is that it never destroys what the user wrote, so tidying away their comments would be the wrong trade.

**Nitpicks** — all taken. The dependency-cycle test (`9e19e7d1`) was worth more than "trivial": termination rests entirely on the `members.insert` guard, and a regression there is an infinite loop in the TUI's poll path, i.e. a hang rather than a wrong answer. The double lookup is collapsed in `2220b9d0` with the guard's cycle-breaking role stated in a comment, so the new test has a named invariant. Refusal-reason assertions tightened in `ffbf2f53`, including a third instance of the same weak pattern sitting between the two you flagged.

Two left alone, with reasons: `write_verified_refuses_text_the_loader_cannot_parse` and `write_verified_refuses_a_digits_only_stack_name` keep their weaker error assertions — they already assert the stronger property (the original file is byte-identical afterwards), and one of them would have to pin text produced inside `serde_yaml` rather than by this module.

116 tests, `cargo fmt --check` and clippy (`--all-features` and `--all-features --all-targets`, `-D warnings`) all clean. Worth flagging for human reviewers: repo CI does not compile or test `workers-dev` — its Rust matrix is driven by directories carrying an `iii.worker.yaml`, and this is a dev tool — so those local gates are the only verification this PR has.

PR #746, opened 2026-08-07
(MOT-4378) feat(provider-github-copilot): Copilot subscription models behind one GitHub sign-in

rohitg00:
Sign in with GitHub once (device code) and the models a Copilot subscription grants appear in the model picker — the same one-login story pi and other mainstream harnesses ship.

## What

New `provider-github-copilot` worker behind llm-router. Wire = OpenAI Chat Completions (provider-openrouter lineage); auth = subscription OAuth with a worker-owned credential lifecycle (provider-openai-codex patterns, no vault dependency):

- **Device-flow sign-in surface** — `provider::github-copilot::login::start` returns the code to enter at the verification URL; `login::poll` stores the GitHub OAuth token in iii-state and refreshes the catalog the moment it lands. Machines already signed in through an editor need no login: read-only import of `~/.config/github-copilot/apps.json` or pi's auth store (`GITHUB_COPILOT_NO_LOCAL_IMPORT=1` opts out; `GITHUB_COPILOT_OAUTH_TOKEN` / `GITHUB_COPILOT_TOKEN` supply tokens directly).
- **Token exchange + refresh** — the long-lived GitHub token is exchanged at `copilot_internal/v2/token` for a short-lived Copilot bearer (~25 min) that also names the API endpoint, so GitHub Enterprise tenants land on their own endpoint automatically. The bearer is cached in-memory, refreshed proactively inside a 2-minute margin, and invalidated when a stream dies with an auth error.
- **Prefixed catalog ids** — Copilot serves several vendors' models under bare ids (`gpt-5.2`, `claude-sonnet-4.6`) that would collide with the sibling single-vendor providers'. Catalog ids are `copilot/<id>`; the prefix is stripped on the wire.
- **Live capability catalog** — admission = `type: chat` + `tool_calls` + plan-enabled (`model_picker_enabled: false` rows fail with "model not supported" and would be dead picker rows). Windows, ceilings, and flags come from the listing's `capabilities` tree. No pricing: subscriptions meter in premium requests, so records carry no per-token cost and `usage.cost_usd` stays unset.
- **Client identity** — the exchange, discovery, and every chat call carry the integration headers the gateway requires, plus `X-Initiator: agent` so agent-initiated turns are billed per Copilot's convention.
- **Errors with subscription semantics** — 401 drops the cached bearer and maps `auth_expired`; 403 (no Copilot access / model not authorized) is `permanent`; numeric `error.code` envelopes win over the transport status.
- **No `count_tokens` surface** — no tokenizer endpoint upstream, and one local tokenizer cannot honestly meter several vendors' vocabularies; the harness falls back to its own estimate.

Carries the transport hardening from the openrouter provider (byte-buffered SSE decode for CRLF framing and split multibyte characters, bounded tool-call indices, credential-redacting Debug).

Also adds the worker to the modules table and `release-workers.yaml`.

## Testing

- `cargo fmt --check`, `clippy --all-targets --all-features -D warnings`, 86 tests: unit (auth chain parsing, exchange reply/cache, device-flow stubs, catalog mapping, admission, wire assembly incl. required headers) + golden wire schemas for all six functions (typed-schema rule enforced).
- Engine-backed integration suite (real engine, real router, stub upstream; the ready-bearer env path short-circuits the exchange so no external API is called): registration + token persistence, catalog-ownership chat streaming, 401 → `auth_expired`, admission + live-metadata reconcile incl. plan-disabled filtering, re-declare on `router::ready`. A suite-wide lock makes plain `cargo test` safe.
- Live verification requires a Copilot subscription sign-in — planned as the rig smoke after merge (`login::start` → device code → catalog fill → picker chat).

Fixes MOT-4378

## Summary by CodeRabbit

* **New Features**
  * Added GitHub Copilot as a supported provider.
  * Added device-flow sign-in and local credential import options.
  * Added subscription-based model discovery with entitlement filtering.
  * Added streaming chat, tool calls, structured output, usage reporting, and request cancellation.
  * Added automatic token exchange, refresh, caching, and provider re-registration.
* **Documentation**
  * Added setup, configuration, authentication, model, endpoint, and troubleshooting documentation.
* **Tests**
  * Added comprehensive coverage for authentication, model discovery, streaming, schemas, errors, and integration scenarios.


rohitg00:
Worked through all 19. Fixed in `4b21b60f` and `62e96290`; the rest are skipped with reasons below.

**Fixed**

- **Probe cost** (`discovery.rs`) — the important one. Verdicts now persist with a 24h TTL, so a model is probed once and answered from cache; the self-healing prune records its verdict too. Verified: cold boot probes 22, warm boot probes 0, same 11 models.
- **Custom `api_url` origin** (`discovery.rs`) — an endpoint not ending in `/chat/completions` fell back to the public host, sending the listing request and its bearer to GitHub instead of the operator's gateway.
- **Bearer cache on 401/403** (`discovery.rs`) — the dead bearer was reused by every stream call until the expiry margin.
- **Env override precedence** (`auth.rs`) — a persisted token masked an explicit `GITHUB_COPILOT_OAUTH_TOKEN`.
- **`model_not_supported` substring fallback** (`errors.rs`) — envelope-only now; the verdict prunes a model, so an echoed phrase removed a working one.
- **Numeric error codes** (`errors.rs`) — application codes and values past `u16` were truncated into HTTP statuses. Also corrected the 403 comment: plan authorization, not moderation.
- **Oversized error body / unbounded SSE buffer** (`upstream.rs`) — both capped, with a transient frame when a stream sends no boundary within 1 MiB. `Stop` now precedes `Done` on the no-`[DONE]` ending.
- **Catalog write race** (`router_client.rs`) — prune's read-modify-write and refresh both go through one lock.
- **Login** (`login.rs`) — `slow_down` is distinct from `pending` and carries the extra wait; the status is a typed enum; the unrecognised reply reports keys only. Token-exchange rejections truncate what they quote.
- **Empty stored token / silent state failure** (`state.rs`, `auth.rs`).
- **Test stub** (`integration.rs`) — reads a complete request before routing.
- **README** — `bash` fence, and `III_URL` is what the CLI actually binds.

**Skipped**

- **`::`↔`__` tool-name encoding** (`wire/names.rs`) — a rig-wide convention shared with the harness decode side and every sibling provider. iii function ids are kebab-case, so `__` does not occur in practice; changing one provider breaks the symmetry.
- **Null arguments on a truncated tool call** (`sse.rs`) — deliberate: `null` fails loudly at the call target, where `{}` would silently invoke with empty arguments. Same in the sibling providers.
- **Per-index tool-call start/end** (`sse.rs`) — the shared SSE state machine; this wire delivers tool-call deltas grouped by index rather than interleaved. Changing the event shape here alone would diverge from every sibling and their goldens.
- **Stream client timeouts** (`upstream.rs`) — the router owns stream budgets (`stream_timeout_ms`, `idle_timeout_ms`) and the pump pings through silence, matching the sibling providers.
- **Empty tool content** (`wire/messages.rs`) — shared-lineage code; the change belongs to every provider at once, not one.

Two findings pointed at the framework rather than this worker, so they are not addressed here: `llm-router`'s custom-schema validator rejects any field containing "token" (`max_tokens` in a provider's own schema trips it), and the console's llm-router form renders only providers that already have a slice in the entry, which is why newly registered providers stay invisible.

rohitg00:
Correction to my previous comment: I listed the two README items as fixed, but had not actually applied them. They are in `9b8d5ca4` now — `bash` on the sign-in fence, and `III_URL` (what `main.rs` binds) instead of `III_WS_URL`. That same doc error exists in the sibling providers' READMEs; left alone here rather than widening this PR.

That commit also rewrites the discovery section, which still claimed a probe costs one token per refresh. It no longer does: verdicts are cached for 24 hours, and a probe that succeeds on a premium model spends a premium request.

On the failing `iii.dev links` check: the flagged URLs live in `prompts/identity.txt`, a verbatim copy of the prompt already on main in the sibling providers, and `https://iii.dev/docs/...` currently returns 403 to any non-browser client (reproduced locally, with and without a browser user-agent). Nothing in this PR broke them — the site is refusing the checker. The last run on main predates the block, so main would fail the same way today.

PR #778, opened 2026-08-12
(MOT-4407) feat(shell): explorer hidden-files filter, bounded tree walk, image preview, folder search, resizable sidebar

rohitg00:
## Problem

Browsing a home directory in the shell explorer shows only dot folders (Slack report from Mike). Two compounding defects:

- `coder::tree` sorts children in byte order, so dot entries come first, and the per-folder cap (default 50) fills with them. Every visible folder lands in the truncated remainder: "partial listing, 572 folders truncated".
- Raising the cap naively is not safe: an unbounded home-directory walk runs past the engine invocation timeout, and flushing the resulting giant frame kills the worker's engine socket (broken-pipe reconnect loop, worker functions unregistered until restart).

While fixing the tree, three adjacent explorer gaps surfaced in testing: images open as mojibake text, path search never returns folders, and the sidebar width is fixed.

## Changes

**`coder::tree`**
- New `include_hidden` input (default `true`, wire-compatible). When `false`, dot entries are omitted before the per-folder cap is counted, so the cap serves visible names. Explicitly rooting at a hidden folder still lists it.
- The walk is now breadth-first under a total node budget (20k nodes). Level order matters: when the budget cannot cover the whole tree it is spent on shallow entries first, so every folder the user can see lists completely and only deep subtrees come back as stubs (`truncated.reason: "max_nodes"`). A depth-first budget returned 7 of a home directory's 123 visible folders.
- Home-directory snapshot on the live rig: 0.28s, 7 MB bounded response, all 123 root folders listed.

**`coder::read-file`**
- New `encoding: "base64"` (single-path full reads): returns the file's exact bytes base64-encoded, for binary payloads. The encoded length is charged against `max_output_bytes`; C210 with `stat`/window/`numbered`.

**`coder::search`**
- Path matching now includes directories. Each `path_match` carries `kind: "file" | "dir"`; content matching still reads files only.

**Shell explorer UI**
- Files tab: hidden entries filtered by default, eye toggle in the sidebar tab row, persisted per workspace tab. Tree request now passes `per_folder_limit: 500` (matching the editor worker's browse call).
- Editor pane: image files (png/jpg/gif/webp/bmp/ico/avif) render as a preview instead of raw bytes.
- Search tab: separate "folders" section; clicking a folder result switches to the files tab, expands the path, and scrolls to it.
- Sidebar: drag handle on the boundary toward the main pane (180 to 560 px, persisted).

## Testing

- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, full test suite (1352 passed, 15 suites), schema goldens regenerated with `UPDATE_GOLDENS=1`.
- New unit tests: hidden filtering (cap interaction, depth, hidden root), node budget (stop + flag, breadth-first ordering), base64 (exact-byte round trip, C210 conflicts, encoded-length budgeting), directory path matches with kind.
- Shell UI: `tsc --noEmit`, esbuild, vitest 83 passed.
- Live-verified on the rig (engine 0.22.1): wire calls against `/Users/rohitghumare` and `/private/tmp`, plus browser passes of all four features in the console explorer.

Closes MOT-4407.

## Summary by CodeRabbit

* **New Features**
  * View image files directly in a read-only preview.
  * Read individual files as exact Base64 data when needed.
  * Search results now include matching folders as well as files.
  * Reveal matching folders automatically in the file tree.
  * Show or hide hidden files with a persistent setting.
  * Resize the sidebar and retain its width between sessions.
  * Configure tree views to include hidden entries.

* **Improvements**
  * Large trees now prioritize shallow entries and report node-limit truncation.
  * File and folder search results are clearly distinguished.

### DIFF
--- a/shell/ui/src/page/EditorPane.tsx
+++ b/shell/ui/src/page/EditorPane.tsx
@@ -10,7 +10,12 @@
 import { CodeEditor, type Host } from '@iii-dev/console-ui'
 import { useCallback, useEffect, useRef, useState } from 'react'
 import { errorMessage, formatBytes } from '../lib/format'
-import { coderReadFile, coderWriteFile, joinPath } from './coder'
+import {
+  coderReadFile,
+  coderReadFileBase64,
+  coderWriteFile,
+  joinPath,
+} from './coder'
 
 /** File-extension → Monaco language id (unknown ids render plain). */
 export function monacoLangFromPath(path: string): string {
@@ -70,6 +75,31 @@ export function monacoLangFromPath(path: string): string {
     truncated body would destroy the tail). */
 export type ReadOnlyReason = 'binary' | 'truncated' | null
 
+/** Extension → MIME for the image preview (SVG stays in the text editor
+    — it's editable markup first). */
+export function imageMimeFromPath(path: string): string | null {
+  const ext = path.toLowerCase().match(/\.([a-z0-9]+)$/)?.[1]
+  switch (ext) {
+    case 'png':
+      return 'image/png'
+    case 'jpg':
+    case 'jpeg':
+      return 'image/jpeg'
+    case 'gif':
+      return 'image/gif'
+    case 'webp':
+      return 'image/webp'
+    case 'bmp':
+      return 'image/bmp'
+    case 'ico':
+      return 'image/x-icon'
+    case 'avif':
+      return 'image/avif'
+    default:
+      return null
+  }
+}
+
 export interface EditorCacheEntry {
   /** What the worker last gave (or accepted) for this file. */
   savedContent: string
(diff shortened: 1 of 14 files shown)


sergiofilhowz (approved):

shell/ui/src/page/EditorPane.tsx
```diff
@@ -70,6 +75,31 @@ export function monacoLangFromPath(path: string): string {
     truncated body would destroy the tail). */
 export type ReadOnlyReason = 'binary' | 'truncated' | null
 
+/** Extension → MIME for the image preview (SVG stays in the text editor
+    — it's editable markup first). */
+export function imageMimeFromPath(path: string): string | null {
+  const ext = path.toLowerCase().match(/\.([a-z0-9]+)$/)?.[1]
+  switch (ext) {
```
sergiofilhowz, line 82: we need to support svg as well

PR #798, opened 2026-08-14
fix(approval-gate): stop duplicate hook bindings deadlocking approvals

### LINKED ISSUE #797: Approval deadlock: approving a held function call re-runs approval::gate (duplicate hook bindings) and parks the turn forever
## Summary

In manual approval mode, approving a held function call in the console chat deadlocks the turn: the approval card disappears but the turn stays `awaiting_functions` forever, with the call re-parked at `approval::gate` and no pending record left to resolve.

## Root cause

Three steps combine:

1. **approval-gate registers duplicate hook bindings.** `retry_hook_bindings` (approval-gate/src/configuration.rs) re-registers the `harness::hook::pre-trigger` binding for `approval::gate` whenever `engine::triggers::info` reports 0 instances. The engine's instance count lags a successful registration, so the retry loop keeps re-binding — observed **171 duplicate pre-trigger instances** (and 171 post-trigger) within ~90s of startup, on a fresh stack. `main.rs` additionally binds directly before starting the loop, guaranteeing at least two.

2. **The harness keeps every instance.** The harness's `HookSet` is keyed by trigger instance id (`harness/src/hooks/mod.rs`), so all 171 identical `approval::gate` bindings are consulted in chain order.

3. **On approval, the chain resumes past only the first holder.** `approval::resolve` → `harness::function::resolve(action="execute")` → `run_pre_trigger(..., resume_after=Some("approval::gate"))`. `chain_slice` skips the **first** matching holder; the **second duplicate re-runs the gate**. The gate's idempotency re-holds (`pending::get` → Some → hold, no new record), the harness re-parks, and the resolve then deletes the pending record — leaving an orphaned park with nothing to resolve. "Nothing happens."

## Evidence

- `engine::triggers::info {id:"harness::hook::pre-trigger"}` → `instance_count: 171` (one gate startup); `harness::hook::post-trigger` → 171.
- Gate startup log: `trigger binding requested` repeated ~0.5s apart from start until "confirmed" (~90s later).
- Trace of an approval: `approval::resolve` ok → `harness::function::resolve` ok → `execute approval::gate` → `{"decision":"hold"}` → `state::set harness_turn` (re-park) → pending record deleted. `turn_resumed: false`.
- E2E repro (fresh stack, current main code): approve → card clears → turn `awaiting_functions` indefinitely.

## Impact

Any deployment using manual approval mode: the first approval after a gate startup storm deadlocks the turn; the operator must restart the harness.

## Suggested fix

- approval-gate: register each hook binding at most once per startup (retry only failed attempts — harness not up yet), and drop the direct pre-bind in main.rs.
- harness: dedupe hook bindings by function_id per hook point (first registration wins) so a misbehaving registrar can never double-run a hook.

faramirezs:
Fixes #797

## Problem

Approving a held function call in console chat deadlocked the turn: the approval card cleared, but the call re-parked at `approval::gate` with no pending record left to resolve — `awaiting_functions` forever.

## Root cause

1. **approval-gate stacked duplicate hook bindings.** `retry_hook_bindings` re-registered the `harness::hook::pre-trigger` binding for `approval::gate` on every loop iteration where `engine::triggers::info` still reported 0 instances. The instance count lags a successful registration, so ~171 duplicate pre-trigger instances (and 171 post-trigger) accumulated in ~90s of startup; `main.rs` also bound directly before starting the loop, guaranteeing at least two.
2. **The harness kept every instance.** `HookSet` keys bindings by trigger instance id, so all 171 identical `approval::gate` bindings entered the chain.
3. **Resume skipped only the first holder.** On approve, `harness::function::resolve(action="execute")` resumes the pre-trigger chain with `resume_after=Some("approval::gate")`; `chain_slice` removes the first match, so the second duplicate re-ran the gate. Idempotency re-held (pending record still present), the harness re-parked, then the resolve deleted the pending record — orphaned park.

## Fix

- **approval-gate** (`configuration.rs`, `main.rs`): each hook binding registers at most once per startup — only failed attempts (harness type not up yet) are retried; the direct pre-bind in `main.rs` is removed.
- **harness** (`hooks/mod.rs`, `hooks/runner.rs`): `HookSet` dedupes by function_id per point (first registration wins, later duplicates dropped with a warning); unregister removes by the stored instance id. Defense in depth — no registrar can double-run a hook.

## Verification

- Fresh stack repro (before): 171 pre-trigger + 171 post-trigger instances; approve → re-hold → turn parked forever.
- After fix: `engine::triggers::info` → 1 instance per hook; single binding registration in gate log; end-to-end console repro — three consecutive approvals each released (`triggering` → `triggered` with result), turn `completed` terminal, `approval_pending` empty.
- Tests: 315 harness lib tests (incl. new `duplicate_function_binding_registers_once`), 161 approval-gate tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup reliability for approval and filesystem monitoring triggers.
  * Prevented duplicate trigger registrations during retries or repeated setup attempts.
  * Ensured failed registrations are retried while successful registrations remain active.
  * Improved cleanup so only the intended trigger is removed.

* **Tests**
  * Added coverage to verify duplicate registrations are ignored and trigger removal behaves correctly.

### DIFF
--- a/approval-gate/Cargo.lock
+++ b/approval-gate/Cargo.lock
@@ -545,12 +545,13 @@ dependencies = [
 
 [[package]]
 name = "harness"
-version = "1.7.0"
+version = "1.8.1"
 dependencies = [
  "anyhow",
  "async-trait",
  "clap",
  "globset",
+ "iii-console-ui",
  "iii-helpers",
  "iii-sdk",
  "jsonschema",
@@ -802,6 +803,18 @@ checksum = "cb68373c0d6620ef8105e855e7745e18b0d00d3bdb07fb532e434244cdb9a714"
 dependencies = [
  "icu_normalizer",
  "icu_properties",
+]
+
+[[package]]
+name = "iii-console-ui"
+version = "0.1.0"
+dependencies = [
+ "iii-sdk",
+ "schemars",
+ "serde",
+ "serde_json",
+ "tokio",
+ "tracing",
 ]
 
 [[package]]
--- a/approval-gate/src/configuration.rs
+++ b/approval-gate/src/configuration.rs
@@ -118,7 +118,8 @@ pub async fn apply_config(cell: &ConfigCell, cfg: WorkerConfig) {
 }
 
 /// Bind the fixed `harness::hook::pre-trigger` hook at worker startup.
-pub fn bind_hook(iii: &IIIClient) {
+/// Returns `true` when the engine accepted the registration.
+pub fn bind_hook(iii: &IIIClient) -> bool {
     match iii.register_trigger(RegisterTriggerInput {
         trigger_type: "harness::hook::pre-trigger".to_string(),
         function_id: "approval::gate".to_string(),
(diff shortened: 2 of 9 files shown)


faramirezs:
Addressed the CodeRabbit finding (leftover instance count suppressed this worker's own registration, leaving the gate detached):

- retry loop now attempts each bind whenever its bound flag is false; readiness gates only the completion condition.
- Added engine-backed regression tests in `approval-gate/tests/hook_binding_retry.rs`:
  - `leftover_instances_do_not_suppress_this_workers_own_registration` — **fails against the previous loop** (verified), passes now.
  - `failed_registration_is_retried_until_the_harness_is_ready`.
- Full approval-gate suite: 180 tests pass.

Refs #797

PR #808, opened 2026-08-14
fix: heal llm-router provider visibility and stale config UI states

anthonyiscoding:
## Problem

The console and harness easily land in invalid states around llm-router and its providers: providers intermittently invisible, and the Workers tab briefly (or permanently) showing the generic schema editor instead of llm-router's config form. Investigation found no single bug — several races and sticky-state paths compound, and every recovery loop failed open and silently.

## Causes fixed

**Providers invisible**
- Router restart marks every restored provider `available=false`; the `router::ready` re-declare fan-out never reaches providers that outlived the restart (engine drops bindings when the type owner disconnects), leaving recovery to each provider's ~3-minute timer.
- The rediscover sweep only nudged *newly* live provider ids, so an already-known provider stuck unavailable was never re-nudged.
- Routing step 2 routed to the unique catalog owner without a `registered()` check, so a stale persisted catalog slice dispatched to a gone provider (`UnknownProvider` at chat time).
- A provider that lost its registration token against a persisted registry was permanently rejected ("bound to another worker") with no recovery path.
- The console gated all `router::*` reads on the **harness** worker's presence, fetched once with no retry, swallowed subscription failures, and never refreshed on reconnect — a router restart after tab load meant an empty picker until a manual page reload.
- Harness classified startup-race dispatch rejections as `Permanent`, killing turns that raced router/provider boot; sessions born during an outage froze the embedded fallback prompt forever.

**Old config UI**
- llm-router registered its injected config form last at boot, after slow engine round-trips, so open consoles painted the generic form first.
- A single failed `console:script` registration meant the generic form for the console's whole session.
- `SKIP_UI_BUILD=1` embedded a stale `ui/dist` silently, and the freshness check ignored the linked `@iii-dev/console-ui` package.
- Open tabs never re-fetched the recomposed entry schema when providers registered later, keeping zero provider cards.

## Changes

- **llm-router**: sweep nudges every live-but-unavailable provider; routing skips unregistered catalog owners; new `router::provider::unregister` operator escape hatch (drops record + catalog slice, recomposes entry schema, emits change events); UI assets register first at boot; build.rs stale-dist guards.
- **console-ui crate**: failed asset-trigger registrations retry with capped backoff.
- **console web**: new `useLlmRouterStatus` presence hook gates the picker on llm-router; provider/catalog reads re-run on router (re)appearance and WS reconnect; unknown-provider events re-read the list; worker lifecycle and `router::provider::changed` invalidate configuration queries including schemas.
- **harness**: topology dispatch rejections (`function_not_found`, `unknown provider`, `provider … unavailable`, `no provider registered`) classify as transient so bounded resume rides out boot races; a steer whose inherited prompt is the pure embedded fallback re-asks the router for the provider identity prompt; clearer budget rejection message.

All fixes stay inside iii primitives (workers, triggers, functions) — no engine changes.

## Testing

- llm-router: 108 unit + 13 engine-backed integration tests, schema goldens regenerated for the new function (`UPDATE_GOLDENS=1 cargo test`, then clean run).
- harness: 315 unit tests (new test covers the transient/permanent classification, including the router's real `provider <id> unavailable` wording).
- console-ui crate: 13 tests.
- console web: `tsc -b && vite build` plus 1206 vitest tests, biome clean on touched files.

## Known ceilings

- `router::ready` can still fire before subscriber replay lands; the boot nudge and sweep now cover that window.
- The console-restart generic-form flash window (~6.5s asset fetch) remains; removing it needs a loader-protocol change.
- A genuine provider-name typo now burns the bounded transient resumes before failing instead of failing instantly.

MOT-4443

## Summary by CodeRabbit

* **New Features**
  * Added provider removal support, including automatic cleanup of associated models and availability updates.
  * Improved provider discovery and model catalog refresh behavior.
  * Added clearer router, provider, and pricing availability status in the console.

* **Bug Fixes**
  * Prevented stale providers from receiving requests.
  * Improved retry handling for temporary router or provider outages.
  * UI registration now retries automatically after startup failures.
  * Restored provider identity when inheriting fallback prompts.

* **Documentation**
  * Clarified pricing error messages and UI registration behavior.

### DIFF
--- a/console/web/src/hooks/use-llm-router-status.ts
+++ b/console/web/src/hooks/use-llm-router-status.ts
@@ -1,0 +1,43 @@
+import {
+  isWorkerPresent,
+  useWorkerPresence,
+  type WorkerPresence,
+} from './use-worker-presence'
+
+/**
+ * Presence probe for the `llm-router` worker. The router owns every
+ * `router::*` RPC the model picker reads (`provider::list`, `models::list`)
+ * and the change triggers it subscribes to, so provider/model UI gates on
+ * THIS worker's presence — gating on the harness starved the picker whenever
+ * the harness was slow or absent while the router was healthy. Thin wrapper
+ * over the generic worker-presence probe.
+ */
+
+/** Engine worker name for the llm-router worker. */
+const LLM_ROUTER_WORKER_NAME = 'llm-router'
+/** Base id for the browser-local handler bound to the `worker` trigger. */
+const LLM_ROUTER_WATCH_FN = 'console::llm-router-watch'
+
+export type LlmRouterStatus = WorkerPresence
+
+/**
+ * @param enabled - only run against the real backend; pass `false` for the
+ *   mock/Storybook backend (treats the router as present so the picker shows
+ *   in isolation).
+ */
+export function useLlmRouterStatus(enabled: boolean): LlmRouterStatus {
+  return useWorkerPresence({
+    workerName: LLM_ROUTER_WORKER_NAME,
+    watchFnId: LLM_ROUTER_WATCH_FN,
+    enabled,
+  })
+}
+
+/**
+ * Whether the router's `router::*` functions are registered and safe to
+ * trigger. False during the initial presence probe and while the router is
+ * absent.
+ */
+export function isLlmRouterAvailable(status: LlmRouterStatus): boolean {
+  return isWorkerPresent(status)
+}
--- a/console/web/src/hooks/use-llm-router-status.ts
+++ b/console/web/src/hooks/use-llm-router-status.ts
@@ -14,9 +14,9 @@ import {
  */
 
 /** Engine worker name for the llm-router worker. */
-const LLM_ROUTER_WORKER_NAME = 'llm-router'
+export const LLM_ROUTER_WORKER_NAME = 'llm-router'
 /** Base id for the browser-local handler bound to the `worker` trigger. */
-const LLM_ROUTER_WATCH_FN = 'console::llm-router-watch'
+export const LLM_ROUTER_WATCH_FN = 'console::llm-router-watch'
 
 export type LlmRouterStatus = WorkerPresence
 
--- a/crates/console-ui/src/lib.rs
+++ b/crates/console-ui/src/lib.rs
@@ -201,8 +201,11 @@ impl ConsoleUi {
     }
 
     /// Register the content function and one trigger per asset; spawn the
-    /// dev watcher when the watch env var is set. Call once, after the
-    /// worker's regular functions are registered.
+    /// dev watcher when the watch env var is set. Call once at boot — order
+    /// relative to the worker's regular functions is free, and registering
+    /// the UI FIRST is preferred: every moment before the console:script
+    /// announcement is a window where an open console renders its generic
+    /// fallback instead of this worker's injected UI.
     ///
     /// Trigger registration failures are warn-logged, not fatal — injected
     /// UI is an accessory, and the SDK replays surviving registrations on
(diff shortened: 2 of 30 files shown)


llm-router/src/registry/register.rs
```diff
@@ -141,3 +144,87 @@ pub fn make_provider_register(
...
+        Box::pin(async move {
+            let id = input.id;
+            if !valid_id(&id) {
+                return Err(RouterError::new(
+                    RouterCode::InvalidRequest,
+                    format!("invalid provider id: {id}"),
+                )
+                .into());
+            }
+            let removed = registry.remove(&id).await.map_err(|e| {
```
ytallo, line 193: This reset removes the record using only a provider id. `metadata.internal` only hides a function from discovery; it does not authorize calls, so any connected worker that knows this id can unregister a provider and then claim the same name with a new token. That bypasses the takeover protection in `RegistryStore::upsert` and turns a buggy worker or an explicitly targeted agent call into provider hijack/DoS. Please put the reset behind an operator-authenticated boundary or capability instead of exposing an unauthenticated bus function.

console/web/src/hooks/use-llm-router-status.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export type LlmRouterStatus = WorkerPresence
+
+/**
+ * @param enabled - only run against the real backend; pass `false` for the
+ *   mock/Storybook backend (treats the router as present so the picker shows
+ *   in isolation).
+ */
+export function useLlmRouterStatus(enabled: boolean): LlmRouterStatus {
+  return useWorkerPresence({
```
ytallo, line 29: `useWorkerPresence` only performs the workers-list read at mount and then relies on lifecycle events. If llm-router is absent for that first probe and is added while the browser socket is disconnected, the add event is lost and `present` remains false indefinitely. `useModelPickerSource` cannot repair this case because its reconnect listener is only installed when `routerAvailable` is already true. Please make this presence probe reconnect-aware so an absent-to-present transition cannot require a page reload.

harness/src/functions/send.rs
```diff
@@ -224,6 +224,34 @@ pub async fn start(deps: &Deps, req: SendRequest) -> Result<StartOutcome, Harnes
     );
     if let (true, Some(prev)) = (inherits_prompt, prev.as_ref()) {
         inherit_prior_system_prompt(&mut options, &prev.options);
+        // A first send that raced a router outage resolved with the EMBEDDED
+        // fallback, and stickiness would freeze that for the session's life.
+        // When the inherited prompt is exactly the pure fallback (no caller
+        // override — those never match the built-in verbatim), ask the router
+        // again: if it answers now, upgrade to the provider identity prompt.
+        if cfg.provider_identity_prompt
+            && options.system_prompt.as_deref()
```
ytallo: Equality with the embedded prompt does not prove that the previous turn used the fallback. A caller can explicitly use the `override` strategy with exactly that text; on the next send it may omit both prompt fields specifically to inherit the resolved prompt verbatim, but this branch will replace that explicit prompt as soon as the router responds. Please persist whether the prior prompt came from the fallback (or otherwise preserve provenance) instead of changing an inherited prompt based only on byte equality.

llm-router/src/registry/rediscover.rs
```diff
@@ -86,14 +87,20 @@ fn provider_id_of(function_id: &str) -> Option<&str> {
...
-    added
+/// Live providers the router does not currently hold as registered AND
+/// available. A newly live provider has no record; one that missed the ready
+/// fan-out (or whose registry was boot-reset) has a record with
+/// `available=false`. Both need the nudge. Registered-and-up providers are
+/// skipped so routine registration churn never spams healthy providers.
+async fn stale_provider_ids(registry: &RegistryStore, live: Vec<String>) -> Vec<String> {
+    let mut stale = Vec::new();
+    for id in live {
+        let up = registry.get(&id).await.map(|r| r.available).unwrap_or(false);
```
ytallo: The registry available flag is not driven by function-lifecycle events: it flips false only after a failed chat dispatch (or at router boot). A provider that disconnects and reconnects while idle therefore remains recorded as available, so this filter suppresses the nudge even though the SDK replayed only functions and not provider application state or catalog. The previous live-set diff classified the reappearing ready handler as newly live. This change reintroduces the multi-minute stale-provider window this sweep is meant to close. Track disappearance and reappearance, or use the event identity, instead of treating the registry flag as connection state.

crates/console-ui/src/lib.rs
```diff
@@ -375,6 +388,32 @@ impl Served {
...
+    iii: Arc<IIIClient>,
+    function_id: String,
+    kind: AssetKind,
+    path: String,
+) {
+    tokio::spawn(async move {
+        let mut delay = std::time::Duration::from_secs(2);
+        loop {
+            tokio::time::sleep(delay).await;
+            match register_asset_trigger(&iii, &function_id, kind, &path) {
```
ytallo, line 416: IIIClient::register_trigger inserts the trigger locally, ignores the send_message result, and unconditionally returns Ok(Trigger). Both the initial call and this loop therefore take the success branch even when the console rejects the registration asynchronously; this task exits after its first attempt and logs a false success. The generic fallback remains sticky for the failure mode this change claims to heal. Retry from an acknowledged result or read-back (or another observable confirmation), not the Result returned by this fire-and-forget API.

anthonyiscoding:
Superseded by #810 and #812

PR #833, opened 2026-08-19
(MOT-4499) feat(console): one keyboard shortcut registry, and shortcuts for workspaces and settings

rohitg00:
## What

The console knew its keyboard shortcuts in three places that could not see each other: a `?` listener in `App.tsx`, a `⌘K` listener in `PaletteHost.tsx`, and a hardcoded five row array in the shortcut overlay that documented whatever someone last remembered to type into it. The palette could not show a row's chord, because nothing knew what the chords were, and the `⌘K` hint was a literal string that read wrong on every keyboard without a Command key.

There is now one registry. Each entry carries what the shortcut does, the group it is listed under, who dispatches it and the chord itself. Three surfaces are derived from it:

- **the global key handler** (`useKeybindings`), which replaces both loose window listeners
- **the shortcut overlay**, generated per group, so it cannot drift from what the keys do
- **the palette**, whose footer hints and action rows read their chords from the same place

Bindings are written once, with `Mod` for the modifier a platform's own menus use, so `⌘K` and `ctrl+K` are one entry rather than two spellings and a regex. Formatting, matching and conflict identity all resolve `Mod` the same way, which is what lets the overlay print `⌘` on a Mac and `ctrl` next to it on Linux from one source.

Two rules become properties of a shortcut rather than of whichever component implemented it:

- **Scope**, so a key can be documented without being claimed. The palette's arrows and `Escape` are listed in the overlay but dispatched inside the palette, because only one of the two is listening at a time.
- **Whether a shortcut survives a focused text field.** This was an inline tag check in the `?` handler and an unstated exception for `⌘K`.

## The shortcuts 

The registry documented four keys, all of them about the palette. The second commit adds the ones the console's own nouns deserved:

| key | does |
|---|---|
| `1`–`9` | select a workspace by position |
| `t` | new workspace |
| `\` | split the workspace into another panel |
| `,` | open settings |

None of them is a chord, and that is the finding rather than the shortcut. Every obvious chord is already spoken for: `Mod`+digit is the tab switcher in Chrome, Safari and Firefox; `Mod+,`, the chord any desktop app would use for settings, opens Chrome's own preferences on macOS; `Mod+[` and `Mod+]` are Back and Forward there too. A page cannot preempt any of them.

Bare keys are free, they are what GitHub and Linear use for the same reason, and the console already had one in `?`. The registry's typing rule is what makes them safe: none fires while the caret is in a field.

`1`–`9` stores one representative chord and fires for every digit, which is also how the conflict check knows the row occupies all nine. Out of range does nothing rather than wrapping.

`MAC_RESERVED` joins the reserved list, for chords the browser claims on one platform and not the other, so the guard stops reporting `Mod+,` as free on Linux, where it is.

## Verification

- `pnpm exec tsc -b --noEmit` clean
- `pnpm test` — 120 files, 1373 tests, all passing (66 new)
- biome clean on the touched files
- `pnpm build` clean
- Live against a local rig, driven with real key events: the overlay lists all three groups with the chords spelled for this keyboard and alternatives separated by "or"; `t` opens a workspace, `1` and `2` select the first and second, `7` with two open does nothing, `\` splits one column into two, `,` opens settings and closes it again; with the composer focused `t` and `,` type their characters and change nothing, while `⌘K` still opens the palette; the palette footer reads `↵ open`, `↑↓ select`, `tab filter`, `esc close` off the registry, and its Keyboard shortcuts row shows `?`

## One thing to look at

Creating or splitting a workspace autofocuses the new pane's "search pages" field, which is existing behaviour and correct on its own. It does mean the bare-key shortcuts stand down immediately after `t` or `\` until you click away, so chaining them needs a deliberate blur. Worth deciding separately whether an empty pane should take focus.

## Notes for review

- `bindings.ts` holds the grammar and knows nothing about what any shortcut does; `registry.ts` holds the list and knows nothing about how a chord is parsed.
- Punctuation keys ignore the shift flag when matching, because `?` is shift and slash on a US layout and unshifted elsewhere. `Shift+/` is still available for anyone who means the physical chord.
- Conflict identity resolves `Mod` per platform, so the test asks the question twice rather than assuming one keyboard.

Fixes MOT-4499

## Summary by CodeRabbit

* **New Features**
  * Added centralized keyboard shortcut handling across the console.
  * Shortcuts now adapt to macOS and other platforms.
  * Updated the command palette and shortcuts dialog to display available bindings dynamically.
  * Added styled key combinations and digit-based workspace selection hints.

* **Bug Fixes**
  * Improved shortcut behavior in text fields and during IME composition.
  * Prevented handled shortcuts from triggering conflicting browser actions.

* **Tests**
  * Added comprehensive coverage for shortcut matching, formatting, conflicts, platform behavior, and registry validation.

### DIFF
--- a/console/web/src/App.tsx
+++ b/console/web/src/App.tsx
@@ -23,6 +23,7 @@ import {
   DialogDescription,
   DialogTitle,
 } from '@/components/ui/Dialog'
+import { KeyCombo } from '@/components/ui/KeyCombo'
 import { Sheet } from '@/components/ui/Sheet'
 import { Wordmark } from '@/components/ui/Wordmark'
 import { EmptyPane } from '@/components/workspace/EmptyPane'
@@ -36,6 +37,7 @@ import {
   useHashRoute,
   type View,
 } from '@/hooks/use-hash-route'
+import { useKeybindings } from '@/hooks/use-keybindings'
 import { useTheme } from '@/hooks/use-theme'
 import {
   type UseWorkspaceTabsReturn,
@@ -45,6 +47,8 @@ import {
   ConversationsProvider,
   useConversationsCtx,
 } from '@/lib/conversations-context'
+import { shortcutPlatform } from '@/lib/keybindings/bindings'
+import { keybindingGroups, resolveBindings } from '@/lib/keybindings/registry'
 import { subscribePanelOpen } from '@/lib/panel-context'
 import { loadEdgeAddDiscovered, saveEdgeAddDiscovered } from '@/lib/storage'
 import { cn } from '@/lib/utils'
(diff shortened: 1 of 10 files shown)


rohitg00:
Both taken, one of them half.

**`Shift+/` matching a bare `/`** — correct, and the comment above it was promising the physical chord while the code ignored shift for every punctuation binding. Fixed as suggested: a binding that names `Shift` now gets the ordinary comparison, while `?` stays layout-independent.

**Applying the same rule to `conflictIdentity`** — not taken, deliberately. `?` matches whether or not shift is down, so it overlaps `Shift+?` rather than being distinct from it: register both and a single keystroke fires both. Matching has to tell them apart; the conflict check has to refuse them as a pair. Comments on both sides now say so, and the test carries the reasoning.

**`Mod+P`** — added. Worth noting it is the odd entry in that list: the page does receive it and could `preventDefault` it, unlike `Mod+W` or `Mod+N`, which never arrive. It is listed because taking Print away from the reader is its own kind of broken, and the guard's job is "do not bind this" either way. The reason is spelled out next to the entry so the list's definition stays honest.

PR #851, opened 2026-08-20
(MOT-4506) feat(iii-directory): create and delete skills, and load system-installed ~/.agents/skills

andersonleal:
Skills reach parity with prompts in the directory worker, and skills installed by external agent tooling under `~/.agents/skills` become visible to the engine.

Fixes MOT-4506. Also carries MOT-4515 (directory console) and MOT-4511 (state worker) — see **Other tickets** below.

## `directory::skills::create` / `::delete`

Skills previously had `list`/`get`/`index`/`update` but no way to author or remove one, so the console's skills tab hid its new/delete buttons. Both functions mirror the existing prompt CRUD (kind-generic registrars in `src/functions/update.rs`).

`create` writes `<skills_folder>/<id>.md` — the global root only — and refuses ids that would land invisible or unreachable:

- **D114** the id already exists (including via the `<id>/index` alias, so `create impeccable` can't collide with an agents-root `impeccable/SKILL.md`), or a file already sits at the target path but is skipped by the scanner
- **D115** the id would be hidden by `filter_unregistered`, reserves an agents-root namespace, or contains a `prompts`/`system-prompts` segment that the classifier would route away from skills scans
- **D113** empty or oversize content (unchanged, shared with `update`)

`delete` removes the file and then any parent directories it emptied, so deleting a namespace can't leave a husk that keeps shadowing lower-precedence roots.

## Read-only `~/.agents/skills` root

A third scan root, `agents_skills_folder`, sits beneath the existing two: **local > global > agents**, shadowed at namespace level. It is deliberately narrow:

- **Read-only.** `update`/`delete` on a skill resolved there return **D116**; `create` never targets it. The guard checks the resolved absolute path, so every id alias hits it.
- **Shallow.** Only `<skill-dir>/SKILL.md` is scanned, one row per installed skill. The `reference/` and `scripts/` payload beside it never floods `skills::list`.
- **Watched only if present.** The fs watcher `create_dir_all`s its roots, so the root is added only when it already exists — a machine without `~/.agents` never gets one created.
- **Filter-exempt by namespace name, not path**, so a manual fork of an agents skill copied into the global root stays visible.

Agents skills are excluded from `skills::index` by provenance (the resolved path), not by name — matching on name would have hidden a real installed worker that happened to share a directory name.

`SkillFrontmatter` gained `name:`, and title resolution is now `title:` → `name:` → body `# H1` → id, which is what makes agents-convention SKILL.md files show a real title.

## Payload skills can find their own files

`directory::skills::get` now returns `path`, the absolute on-disk file; its parent is the skill's base directory. The console appends a `Skill base directory: …` line when it freezes a skill into a session addon, at both freeze sites (the addon picker and `/skill:<id>` expansion) via one shared helper.

This closes a measured failure: in live session `console-1f62d4cf`, a skill whose body says to run `scripts/context.mjs` had no way to say where that lives, so the model probed `.agents/skills/…` relative to the project, got ENOENT, and silently degraded to prose-only — skipping the script, ~20 reference playbooks, and its subagent definitions. Body-only injection is preserved for workers that send no `path`.

## Console UI

- Skills tab gets create/delete wired to the new functions, with a slash-allowing `namePattern` on creates only (on update the frontmatter title is a display field, and the slug gate was blocking saves of human-readable titles)
- `directory::search_functions` renders its result card inline in chat (`metadata: { display: true }`) instead of hiding it behind "show raw request and response"
- Config form brought in line with the #837 house style: sentence case, shared `Input`/`Chip` primitives, no panel-wide mono, focus ring restored. The uppercase CSS was defeating #837's own text fix — it still rendered `RESTART REQUIRED`, and mangled the product name to `III-DIRECTORY`.

## Other tickets on this branch

- **MOT-4515** — unsaved skill/prompt drafts were lost whenever you switched console tabs (the console unmounts the page; the draft lived only in React state). Drafts now mirror to localStorage under the existing per-tab, per-collection key and restore on mount. The write/clear/keep decision is a pure module so it is testable without a DOM — the non-obvious case being that storage is *kept* while a baseline load is in flight, since clearing there would destroy the work just restored.
- **MOT-4511** — ⚠️ **behaviour change in the `state` worker**: the kv store now defaults to `file_based` persistence, with a flush on shutdown and on SIGTERM. Harness turn records live in the state worker, so with the old in-memory default a state restart silently dropped each session's model/provider/options. Existing deployments that *want* in-memory must now say so explicitly. Committed as `feat(state)!`. Happy to split this into its own PR if you'd rather land it separately.

## Verification

- `iii-directory`: build + `clippy --all-targets --all-features -D warnings` clean, **354** lib tests, `--manifest` includes the new field
- `state`: build + clippy clean, **134** lib tests
- UI: `pnpm build` (tsc + esbuild) + **30** tests; console `tsc` clean + **58** tests across touched suites
- Live-verified against a running engine: `directory::skills::get { id: "impeccable" }` returns `path: /home/anderson/.agents/skills/impeccable/SKILL.md`

Rebased onto the SDK 0.23.0-rc.2 migration (#604), which moved both `iii-directory` and `state` off 0.21.8. One conflict, in `state/tests/e2e_state.rs`, where that migration's `CONFIG_ID` → `config_id()` rename met this branch's `StateConfig::default()` → `in_memory_config()` change; both were kept, since the default is now file_based and that test needs in-memory.

Not gated on `cargo test --test bdd` — 25 pre-existing environment-dependent failures on main, unchanged by this branch.

## Summary by CodeRabbit

* **New Features**
  * Browse system-installed skills from a read-only skills directory.
  * Create, update, and delete skills with improved validation and visibility rules.
  * Preserve unsaved skill and prompt edits when navigating away.
  * View and copy session IDs from context popovers.
  * Persist state to disk by default for improved restart continuity.
  * Search results can now appear directly in chat.

* **Bug Fixes**
  * Ensure skill prompts include their correct base directory when invoked.

* **Documentation**
  * Added guidance for issue tracking, triage labels, domain documentation, and skill management.


andersonleal:
Worked through all ten CodeRabbit findings — eight applied in `b92d7de4a` and `810d1612b`, two skipped with reasons.

### Applied

**state worker (`b92d7de4a`)** — all three flush findings were real, and each loses exactly the write that the `file_based` default exists to preserve:

- *Serialize flushes.* Confirmed: the periodic loop and `flush()` both `drain()` the whole dirty map with no mutual exclusion, so the shutdown flush could see it empty while the loop's write was still in flight, return, and let process exit cancel that write. Now a shared async mutex spans drain → persist → requeue on both paths.
- *Don't clobber newer intent on requeue.* Confirmed, and this one silently destroys data: a failed `Delete` requeued with `insert` overwrites an `Upsert` a concurrent `set` queued while the write was in flight, so the next flush deletes a scope the store still holds. Now `entry().or_insert()`. Covered by a test.
- *Propagate flush failures.* Confirmed — `BootHandle::shutdown` already had a `warn` branch for a failed final flush, but it was unreachable because `KvStoreAdapter::flush` discarded the result. `flush_dirty` now attempts every scope, then reports how many failed. Making it a `Result` turned the discard into a compiler error, which is a nice property. Covered by a test that plants a directory where the index file belongs.
- *Quiesce before shutdown.* Took the cheap 80% rather than the proposed admission control: `shutdown_async` clears the SDK's `running` flag, and its receive loop is gated on that (`iii.rs:1707`), so simply calling it **before** `boot.shutdown()` stops new invocations from being dispatched behind the flush. Requests already executing stay a race — the SDK exposes no way to await handlers, and bespoke in-flight tracking across every handler is out of scope for this PR.

**directory / console / harness (`810d1612b`)**

- *Windows separators in `skillBodyWithBaseDir`.* Valid — and there was a second defect underneath it: with no separator at all, `lastIndexOf` returns `-1` and `slice(0, -1)` yields the whole path minus its last character, which the guard below let through and presented to the model as a directory. Now cuts on either separator and falls back to body-only. Tests for a Windows path and for the no-separator case.
- *Document external events on the trigger.* Valid, and the same omission was in all three descriptions, so all three were fixed rather than just the flagged one. The prompts description was also missing `delete`, which it has fired since the prompt CRUD landed.
- *README anchor.* Valid — `#layout` vs the heading's `#on-disk-layout`.
- *Session-id truncation.* Valid: a flex item won't shrink past its content without `min-width: 0`, so the id pushed the copy button out of the 300px popover. Added to the id only, matching the sibling `.harness-ui-pop-model` that already pairs the two; the row itself is a column-flex item, where the automatic minimum applies to the block axis.

### Skipped

- **`docs/agents/domain.md` fenced-code languages** — moot. Those files were dropped from the branch after the review ran, so the file no longer exists here.
- **Re-watch the agents root if it appears later** — genuine, but not worth the machinery. There's no cheap correct version: `spawn_fs_watch` `create_dir_all`s its roots and this worker must never materialize `~/.agents`, and the only ancestor guaranteed to exist is `$HOME`, which can't be watched recursively. More importantly the blast radius is small — the watch is a doorbell, not a ledger, and every read re-scans disk, so a skill installed into a freshly created root is still **served** correctly; only live refresh waits for a restart. Documented in the README rather than left implicit.

PR #925, opened 2026-08-25
Run isolated Harness E2E campaign groups

ytallo:
## Summary

- add the campaign v3 dispatch contract while preserving legacy v1/v2 executions
- expand one campaign into isolated group jobs with `fail-fast: false`
- route common groups to exact ephemeral stacks and fault groups to the protected trusted runner
- preserve native `results.json`, `manifest.json`, and `observation.json` bytes and digests
- aggregate partial group outcomes in a final exact Harness runner job

## Validation

- 229 workflow/script tests passed
- 3 shell subtests passed
- shell syntax, workflow YAML, and diff checks passed

## Dependency

Depends on iii-hq/harness-e2e#24. This is the second PR in the Harness -> Workers -> Release Control rollout.


ytallo:
Superseded by iii-hq/harness-e2e#24 and iii-hq/release-control#38. Campaign v3 execution now belongs to Harness E2E; Workers remains only a versioned source of components in the tested stack. Historical v1/v2 executions continue using the existing workflow on Workers main.

PR #940, opened 2026-08-26
refactor(console): trace views ride the trace trigger, not span streams

ytallo:
Ref: [MOT-4540](https://linear.app/motia/issue/MOT-4540/investigate-stuck-traces-traces-that-never-complete) · Engine counterpart: iii-hq/iii#2088

The engine's devtools span streams (`iii:devtools:trace-rows` / `trace-spans` / `all-spans`) are removed by iii-hq/iii#2088 — they serialized every 300ms window and forced this frontend to re-implement list semantics client-side (root-wins merge, tag backfill, dropped-frame self-heal). All three trace surfaces now ride the coalesced `{trace_ids}` tick of the `trace` trigger the timeline strip already used, and re-run their own seeded, filtered queries (notify-then-query):

- **list** — tick → debounced invalidate of `traces`/`traceGroups`/`traceGroupMembers`; the append cache, tag patching and backfill machinery disappear (a refetch always carries tags and cannot drift from the server's filter semantics)
- **detail** — silent reload of the open trace when its id arrives in a tick; one request in flight with a trailing rerun
- **masthead strip** — debounced re-seed (REPLACE semantics, same read as the initial seed)

Pause, tab-hidden, reconnect-reseed and hover-hold behavior are unchanged. `mergeDetailSpan` and the stream extractors go away with their tests; the activity feed moves to `lib/traces-activity.ts`.

**Validated live** against an engine running iii-hq/iii#2088 (streams gone): ticks delivered to all three subscribers, list refetches observed carrying the new traces (`total`/first-row advancing per tick), zero requests while idle. Web suite: 1629 tests green; tsc + biome clean on touched files.

**Deploy order**: this must ship before (or with) iii-hq/iii#2088 — a console running the old stream subscriptions gets no live trace updates against an engine without the streams (list still seeds on load/reconnect; nothing breaks hard).

https://claude.ai/code/session_01LoPzhwFhAzxsrFnRqEEga6

## Summary by CodeRabbit

* **New Features**
  * Added navigation from trace details directly to related chat messages.
  * Chat messages now open, center, and briefly highlight selected transcript entries, including grouped messages.
  * Traces automatically scope to the active conversation, with dismissible filters and contextual empty states.
  * Added “show all sessions” access and progressive loading for trace details.
  * Improved trace refresh behavior with activity-based updates.
* **Bug Fixes**
  * Improved handling of hydrated transcripts, stale navigation requests, and session-specific trace links.
* **Tests**
  * Added coverage for trace-to-chat navigation, filtering, activity updates, and transcript anchoring.


guibeira (commented):
Do you mind to solve the conflicts?

PR #1018, opened 2026-08-30
fix(mcp): remove stale dependency on the removed skills worker

### LINKED ISSUE #1017: mcp is uninstallable from the registry: stale dependency on the removed `skills` worker
### Summary

No version of `mcp` can currently be installed from the worker registry: every published release (0.5.7–0.5.9) declares a dependency on a worker named `skills`, which no longer exists. Its functionality was folded into `iii-directory` in May, the `skills/` tree was deleted from main, and the worker was withdrawn from the registry — but the mcp manifest was never updated. Registry resolution fails for all of them, so `iii compose build` / `iii worker add mcp` always ends in `PACKAGE_NOT_RESOLVED`.

### Repro

```
$ curl -s -X POST https://api.workers.iii.dev/resolve \
    -H 'Content-Type: application/json' \
    -d '{"worker":"mcp","version":"0.5.9","target":"aarch64-apple-darwin"}'

{"error":{"available":[],"code":"worker_not_found","dependency_path":"mcp -> skills",
 "fix":"Publish 'skills' or remove it from the dependency list.",
 "message":"Worker 'skills' was not found in the registry.",
 "requested":["^0.2.1"],"scope":"worker","worker":"skills"}}
```

Identical for 0.5.7 and 0.5.8, and the registry carries no older mcp versions (`available: ["0.5.9","0.5.8","0.5.7"]`), so there is no installable fallback.

### Root cause — timeline

- **2026-05-07** — #98 lands the `skills` worker (filesystem-backed skills + prompts); bumped to v0.2.4, published to the registry.
- **2026-05-12** — #124 (`feat: iii directory`) deletes `skills/` from main: the functionality lives on as `iii-directory` (`directory::skills::*`, `directory::system-prompts::*`). The `skills` worker disappears from the registry.
- `mcp/iii.worker.yaml` keeps `dependencies: skills: "0.x"` throughout, so the 0.5.x releases cut by the new release train (#997, 2026-08-29) all reference the removed worker.

Two stale links remain in the mcp tree:

1. **Manifest** — the `dependencies` block above; this is what blocks registry resolution.
2. **Runtime bridge** — `mcp/src/skills_bridge.rs` still delegates MCP `resources/*` and `prompts/*` to `skills::resources-list` / `skills::resources-read` / `skills::resources-templates`. No worker registers those IDs any more (`iii-directory` registers `directory::skills::*` instead). The bridge degrades gracefully (`delegate_or_empty` returns empty lists on function-not-found), so tools keep working — but MCP resources/prompts can never be populated unless the bridge is repointed.

### Suggested fix

1. Drop the `dependencies` block from `mcp/iii.worker.yaml` and cut a release (PR attached).
2. Follow-up: either repoint `skills_bridge.rs` at the `directory::skills::*` surface, or have `iii-directory` register `skills::*` resource aliases — otherwise `resources/list` stays permanently empty post-install.

ruttybob:
Fixes #1017

## What

Removes the `dependencies` block (a single stale entry, `skills: "0.x"`) from `mcp/iii.worker.yaml`.

## Why

The `skills` worker was removed from main and withdrawn from the registry when its functionality was folded into `iii-directory` (#124, 2026-05-12), but the mcp manifest kept referencing it. Every mcp release currently in the registry (0.5.7–0.5.9) therefore fails registry resolution — `iii compose build` and `iii worker add mcp` both end in `PACKAGE_NOT_RESOLVED: Worker 'skills' was not found in the registry` — and there is no older installable fallback, so the mcp worker is currently uninstallable for everyone.

With the dependency dropped, the worker resolves and installs again. Runtime behaviour is unchanged for tools; `resources/*` and `prompts/*` already degrade to empty lists when no delegate registers (`skills_bridge.rs` `delegate_or_empty`), since `iii-directory` exposes the `directory::skills::*` surface instead.

## Follow-up (not in this PR)

`skills_bridge.rs` still delegates to `skills::resources-*` function IDs that no worker registers. To make MCP resources/prompts populate again, either repoint the bridge at `directory::skills::*` or have `iii-directory` register `skills::*` aliases.

## Summary by CodeRabbit

* **Chores**
  * Updated the worker configuration by removing an outdated package dependency declaration.

### DIFF
--- a/mcp/iii.worker.yaml
+++ b/mcp/iii.worker.yaml
@@ -7,6 +7,3 @@ license: Apache-2.0
 bin: mcp
 tags: [mcp, model-context-protocol, bridge, http, agents]
 description: MCP 2025-06-18 Streamable HTTP bridge for the iii engine.
-
-dependencies:
-  skills: "0.x"


guibeira:
Hey @ruttybob , thanks for let us know, let me test it.

PR #1050, opened 2026-09-02
(MOT-4621) fix(console): calm the traces list under sustained activity

ytallo:
Fixes MOT-4621

## Summary

- add `lib/refetchCoalescer.ts`: one refetch in flight per surface (list, group aggregate, timeline strip), exactly one trailing rerun for ticks that land mid-flight, a 1s floor between starts, `cancelRefetch: false`
- hide span-filtered rows whose composition verdict is still unknown until a read or the feed proves visible work; keep rows that are still running (`pending` status or a pending snapshot in the feed), record no verdict for a trace a read finds running and read it again once it settles; drop the "ended within 10s" liveness slack
- keep the span-filter selection identity stable across producer-hidden refetches (`structuralSharing` on equal id sets)
- bound the strip's summary seed to 120 traces (was 500)
- keep grouped member rows on screen while the count-keyed refetch loads (`keepPreviousData`); stop cutting a row's arrival flash short when the next answer lands

## Context

Reported from a three-agent orchestration run: new traces "jump a lot" and rows keep switching states. Reproduced with four concurrent harness turns (~2 turns/s, ~40 spans each) against a local engine, watching the page with Playwright (row sampling every 100ms + WebSocket frame capture). Four independent causes, all in the console:

1. **Every `trace` trigger tick invalidated the list, the group aggregate and the strip seed.** react-query cancels the fetch already in flight on `invalidateQueries` (`cancelRefetch` defaults to `true`) — client-side only, the engine keeps executing the abandoned scan. Under sustained ticks the scans piled up (six `engine::traces::list` in flight at once, p50 1.9s, max 5.9s) and every answer was discarded before it landed: the list froze while activity lasted, then jumped to the final state in one batch, with dozens of rows appearing, flashing and changing status at once.
2. **Rows with a hidden root were shown while their verdict was unknown.** Every hidden bookkeeping trace (`harness::state::list`, `*::on-functions-change`, `session::messages`) flashed into the live list and vanished when its composition read landed, shifting everything below it.
3. **The span-filter selection identity changed on every `engine::functions::list` refetch** (a fresh Set with equal contents), resetting the verdict cache and blinking rows out and back in.
4. **The strip re-seeded 500 trace summaries per tick** (~157KB, ~8s per call under load) — the engine loads every span of those traces to aggregate them, the single heaviest query on the connection.

## Validation

Same load, flat list, 60s of observation, before → after:

| metric | before | after |
|---|---|---|
| concurrent `traces::list` calls in flight | 6 | 1 |
| rows that vanished and came back | 2 | 0 |
| reorders / re-flashes | 0 / 0 | 0 / 0 |
| UI updates during the load | frozen until activity paused, then one batch | steady, every ~1–2s |

The grouped view now updates counts continuously under the same load (44 count changes in 55s vs. none before). Unit tests: 207 pass in `TracesV2` + `lib` (new coverage for the coalescer, `mergeFetchedVerdicts`, `pendingTraceIds`, `mergeNewTraceIds`, `sameHiddenIdSet`); `pnpm typecheck` clean; biome clean on the touched files. Two failures in `MessageList.test.tsx` are pre-existing on `main`.

## Notes

- Live updates depend on the engine delivering the `trace` trigger in the console's namespace — engines older than iii-hq/iii#2101 (2026-08-28) only deliver it in `default`, so a namespaced console never refreshes; that is engine-side and unrelated to this change.
- Inherent to the data model and left alone: the `iii.tag.message` row label arrives after the root, so a turn row relabels from `execute harness::send` to the message text once the tag-bearing span closes. Stamping the tag on the root span would be a harness change.

https://claude.ai/code/session_01Ass3crjJFx4s776BYMpdxw

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented visible trace rows from disappearing and reappearing during refreshes.
  - Improved filtering of unknown, pending, running, and error traces.
  - Added deferred checks and retries for composition-based filtering.
  - Preserved new-trace indicators and hidden-function filtering across updates.
  - Stabilized the ordering of grouped trace results.

- **Performance**
  - Smoothed activity-driven refreshes and reduced summary loading work.
  - Preserved existing results while grouped trace data refreshes.
  - Reduced unnecessary updates when hidden-function selections are unchanged.

### DIFF
--- a/.github/scripts/tests/test_worker_dependency_compatibility.py
+++ b/.github/scripts/tests/test_worker_dependency_compatibility.py
@@ -37,7 +37,9 @@ DEPENDENCY_RANGES = {
     "queue": "^0.21.5",
     "session-manager": "^1.0.13",
     "shell": "^0.11.9",
-    "skills": "0.x",
+    # `skills` left this table with the worker itself: its functionality was
+    # folded into iii-directory and the last consumer (mcp) dropped the
+    # dependency in #1018, so a range here would fail the consumer check.
     "state": "^0.22.2",
 }
 
--- a/console/web/src/lib/trace-hidden-functions.test.ts
+++ b/console/web/src/lib/trace-hidden-functions.test.ts
@@ -1,0 +1,19 @@
+import { describe, expect, it } from 'vitest'
+import { sameHiddenIdSet } from './trace-hidden-functions'
+
+describe('sameHiddenIdSet', () => {
+  it('treats equal contents as the same set, whatever the instances', () => {
+    expect(sameHiddenIdSet(new Set(['a', 'b']), new Set(['b', 'a']))).toBe(true)
+  })
+
+  it('sees any difference in membership', () => {
+    expect(sameHiddenIdSet(new Set(['a']), new Set(['a', 'b']))).toBe(false)
+    expect(sameHiddenIdSet(new Set(['a', 'b']), new Set(['a', 'c']))).toBe(
+      false,
+    )
+  })
+
+  it('never matches a missing previous value', () => {
+    expect(sameHiddenIdSet(undefined, new Set())).toBe(false)
+  })
+})
--- a/console/web/src/lib/trace-hidden-functions.ts
+++ b/console/web/src/lib/trace-hidden-functions.ts
@@ -37,6 +37,21 @@ function parseTraceHiddenIds(res: unknown): ReadonlySet<string> {
 }
 
 /**
+ * Same ids, whatever the Set instances. A refetch parses a fresh Set even
+ * when nothing changed, and react-query's structural sharing only reuses
+ * plain objects and arrays — callers that key memos (and the span-filter
+ * verdict caches) on the set's identity use this to keep the old instance.
+ */
+export function sameHiddenIdSet(
+  previous: ReadonlySet<string> | undefined,
+  next: ReadonlySet<string>,
+): boolean {
+  if (!previous || previous.size !== next.size) return false
+  for (const id of next) if (!previous.has(id)) return false
+  return true
+}
+
+/**
  * Function ids registered with `trace_hidden: true`. Resolves to the empty
  * set on any failure — the traces page then simply hides nothing by default.
  */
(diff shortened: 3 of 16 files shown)


ytallo:
## Evidence video: before vs. after under the same load

▶ https://claude.ai/code/artifact/a168dfc8-72ce-4e84-b43b-a7cf0c512160

Side-by-side recording (1800×1000, 80 s): **left** is `main` at 1143760d7, **right** is this branch, both served from Vite dev against the same local engine and embedded in one Playwright page while four concurrent harness turns (~2 turns/s, `gpt-5.4-nano`, one session per lane) run. The counter above each pane is how many times its visible row set or any row status changed, and how long ago.

What to look for:

- **Before:** rows land in bursts of four to eight, all flashing at once, after seconds of nothing; hidden bookkeeping traces (`execute session::get`, `directory::on-functions-change`, `harness::state::list`) enter the list and vanish moments later, shifting everything below them.
- **After:** one or two new rows at a time, each flashing once, on a steady cadence; bookkeeping traces never appear.

The page also carries the measured numbers from the PR description and the recording setup. The raw `.webm` (5 MB) was delivered to the author for native attachment; GitHub's API does not accept media uploads on comments, hence the link.

ytallo:
Two follow-up commits landed on this branch while validating the stacked "hold while reading" PR, both fallout of the new *unknown hides* rule in `useSpanFilteredTraceRows`:

- `accb0afab` — a trace that paged out and back in stayed hidden for good: its verdict was pruned on the way out, but the read bookkeeping still said it had been read. Read states now live only while the trace is listed, a trace found running is re-read once its row settles, and a failed read retries after a backoff.
- `753516a43` — the composition read asked for up to 10 000 full spans of 20 traces; under load that outgrew the transport's ~16 MiB delivery cap and never arrived (no error), so the rows waiting on it never showed (page 1 kept 10 of 50 rows after a visit to page 2). Rows are now judged by bounded per-trace probes (first 80 spans, four in flight), which are complete for the bookkeeping traces this path hides and can only prove visibility for a fat one.

Re-validated under the same load after both: page 2 fills to 42–47 visible rows within ~10 s and page 1 comes back whole.

PR #1053, opened 2026-09-02
(MOT-4640) perf(harness): advertise MessageInput as a plain string

andersonleal:
Closes [MOT-4640](https://linear.app/motia/issue/MOT-4640). Sub-issue of MOT-4636; follows #1047 (MOT-4638).

## Why

`SpawnRequest.task` and `SendRequest.message` are typed `MessageInput`, whose `Message(Box<AgentMessage>)` arm dragged the **entire transcript type tree** into both *input* schemas: 14 definitions (4,196 chars). Seven of them — `AssistantMessage`, `FunctionResultMessage`, `Usage`, `StopReason`, `ErrorKind` and the assistant/function-result role tags — describe shapes `normalize_message` rejects at runtime, so the contract was advertising payloads the door refuses. In the corpus, 27 of 27 real `harness::spawn` calls passed `task` as a plain string.

## Change

`MessageInput` gets a hand-written `JsonSchema` that advertises `string` only (`is_referenceable = false`, so it inlines under each field's own doc). One type, both functions fixed.

Deserialization is **untouched**: the structured `user`/`custom` arm stays a compatibility path for programmatic callers — the console sends `role: user` messages carrying `#file(...)` blocks (`console/web/src/lib/backend/harness-send.ts`). Field docs on `SendRequest.message` / `EditQueuedRequest.message` no longer promise the object form.

## Measured

Request schemas on the wire, read via `engine::functions::info` from an ephemeral engine running this branch's release binary:

| function | before | after |
| -- | -- | -- |
| `harness::spawn` | 12,184 | **7,718 (−37%)** |
| `harness::send` | 11,019 | **6,483 (−41%)** |

Both now carry zero transcript definitions. The golden diff (`harness/tests/golden/schemas/harness.{spawn,send}.json`, −884 lines) is the reviewable deliverable.

## Tests

- `message_input_schema_is_a_string_and_hides_the_transcript_tree`: pins `type: string` on both fields, asserts none of the ten transcript definitions appear in either request schema, checks the 27 corpus spawn payloads' `task` field against the new schema (one payload was already invalid on `display.icon` before this change, so only `task` verdicts are pinned), and round-trips a structured `user` message through `normalize_message`.
- Existing `assistant_message_is_rejected` still covers the `InvalidMessageRole` path.
- `cargo fmt`, `cargo clippy --all-features --all-targets -D warnings`, full `cargo test` green; goldens regenerated with `UPDATE_GOLDENS=1`.

Not exercised live: a full `harness::send` turn (needs the whole provider stack). The wire contract check above is what the model sees.

https://claude.ai/code/session_01EUoLR66baA2QL7x6fH72bT

## Summary by CodeRabbit

* **Documentation**
  * Clarified that send and spawn message fields represent user-provided text.

* **Bug Fixes**
  * Simplified published request schemas so message and task fields are represented as plain strings.
  * Prevented internal transcript message structures from appearing in public schema definitions.
  * Preserved compatibility with existing structured message payloads.


andersonleal:
## Live verification on the `my-project` compose stack (2026-09-02)

Swapped the running harness for this branch's release binary (same env, same engine), then ran probe session `mot4640-live-1` with model `codex/gpt-5.6-terra`.

**Wire contract** (`engine::functions::info`, same engine, before → after the swap):

| function | before | after | transcript defs |
| -- | -- | -- | -- |
| `harness::spawn` | 12,184 (24 defs) | 7,718 (10 defs) | 0 |
| `harness::send` | 11,019 (22 defs) | 6,483 (8 defs) | 0 |

**Model-visible copy** (what the parent actually received in its transcript after #1047's compaction):

| | `harness::spawn` contract |
| -- | -- |
| yesterday, #1047 only (`mot4638-live-2`) | 11,545 bytes |
| today, #1047 + this PR (`mot4640-live-1`) | 7,746 bytes (−33%) |

**Turn**: `harness::send` with a plain-string `message` → the model fetched the spawn contract, called `harness::spawn` with `task` as a string, and completed with `SPAWNED mot4640-live-1-child`. The child session was created and ran its own turn end to end. (It ended with a policy denial on `state::set` because my probe gave the child an allow list the parent did not hold — a probe mistake, unrelated to this change.)

PR #1097, opened 2026-09-04
feat(console): open the function document on its contract

### LINKED ISSUE #223: [iii console] Display Input and Output Schemas in the Functions Detail Pane
Currently, when selecting a function under the Functions tab in the iii console, the right-hand pane displays the function name, a short description, and a text area to input JSON for invocation. However, there is no immediate visibility into what properties the function expects (Input Schema) or what it returns (Output Schema).

tacoda:
Selecting a function in the Functions catalog landed on the invoke editor: a name, a description, and a JSON box, with nothing on screen about the fields that box expects or the shape that comes back. The input and output schemas were a tab click away, and the contextual rail only answered "defined" or "none".

Fold the Input and Output tabs into Overview and open the document there, so the contract is the first thing on the work surface. Overview now renders the input schema, the output schema, and any registered metadata; the tab strip drops from six entries to four; `run function` in the identity head still switches to the editor in one click.

No new styles: the section labels and the overview column reuse the existing `console-catalog-field-label` and `console-catalog-overview` rules, and the field tables are the same `SchemaTable` the two removed tabs rendered.

Verified against a live engine (iii 0.23.0) with the `shell` worker registered: selecting `shell::fs::ls` opens on Overview with the `LsRequest` and `LsResponse` field tables, and the tab strip reads Overview / Trigger / Triggers / Activity.

After overview:
after-overview

Metadata function:
metadata-fn

Trigger tab:
trigger-tab

**Only validated on iii 0.23.0** - Needs testing on older versions for backward compatibility.

I verified some dead code is actually dead and removed it. This works on 0.23.0, but I'm suspicious older versions rely on this api.

Closes #223 

## Summary by CodeRabbit

* **User Experience**
  * Selecting a function now opens its Overview tab by default.
  * The Overview displays the function contract, including input schema, output schema, and metadata.
  * Removed the separate Input and Output tabs from the function view.
  * Updated interface copy to clarify that the Overview represents the function’s contract.

* **Documentation**
  * Clarified that schemas that cannot be represented as fields display their raw JSON schema instead of an empty table.

### DIFF
--- a/console/ui/src/catalog/FunctionsPage.tsx
+++ b/console/ui/src/catalog/FunctionsPage.tsx
@@ -3,7 +3,12 @@
  * function on the bus — each row led by the `ƒ` tile, grouped by the worker
  * that registered it — and a workspace that is always present: a hero when
  * nothing is selected, the function document (breadcrumb, identity head,
- * overview/invoke/input/output/triggers/activity tabs) when one is.
+ * overview/invoke/triggers/activity tabs) when one is.
+ *
+ * The document opens on the overview, and the overview is the contract:
+ * input schema, output schema, metadata. Selecting a function used to land
+ * on the invoke editor, which showed a name, a description, and a JSON box —
+ * nothing about the fields that box wanted.
  *
  * Live, never polled. `engine::functions-available` fires whenever functions
  * are registered or unregistered, so a worker connecting or dying is visible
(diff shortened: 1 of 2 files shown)


guibeira (approved):
Hey @tacoda , nice to see you again on the PRs. Thanks again for the contribution.

PR #1120, opened 2026-09-08
(MOT-4722) fix(iii-directory): send the search guidance preambles once per session

andersonleal:
## Why

`directory::search_functions` re-sent `SEARCH_GUIDANCE` (~1 KB) and `SEARCH_INSTALL_WORKFLOW` (~3 KB) on every call. In the Linkly agentic e2e run (MOT-4717, finding A9) the agent searched 4-6 times per turn, so the same text landed in the transcript a few dozen times per session. `SessionRegistry::split` only deduplicated function ids, never the preamble.

## What

- `SessionRecord` gains `preambles: HashSet<&'static str>`; `SessionRegistry::first_send(session_id, preamble)` is true once per session per block.
- `search_functions` sends the full `SEARCH_GUIDANCE` / `SEARCH_INSTALL_WORKFLOW` the first time and a one-line pointer after that (`SEARCH_GUIDANCE_REPEAT`, `SEARCH_INSTALL_WORKFLOW_REPEAT`).
- Reset rides the same paths that wipe `delivered`: a catalog fingerprint change, or an all-repeat query (the existing compaction-recovery path), so the agent never loses the rules for good.
- No session id in baggage → every call is a first call (fail-open, same stance as the candidate dedupe).
- `installed_compose_add_carries_installation_guidance_without_installable_candidates` updated: it pinned that the second call in a session repeats the full workflow, which is exactly what changes here. New unit test for `first_send` and its reset points.

## Not in this PR

Skills of non-installed workers are invisible to `directory::skills::*` because the registry's `/w/{name}/skills` publishes `[]` for e.g. `http` / `rbac-proxy` — a registry publish gap, tracked on MOT-4722.

## Tests

- `SKIP_UI_BUILD=1 cargo test`: unit 441/441 (3 ignored). `cargo clippy --all-targets` clean.
- BDD (`tests/bdd.rs`) runs against the live engine + registry and is environment-bound here: the run with this change failed 18 steps, all of them also failing (among 32) in a stashed baseline run — registry downloads (`D310 … has no published skills bundle`, `.invalid` git hosts), skills-index contents of the live stack. No new failures.

Part of MOT-4717 · closes MOT-4722

https://claude.ai/code/session_01PvtXK7JgHHNrG192afbRAk

### DIFF
--- a/iii-directory/src/functions/search.rs
+++ b/iii-directory/src/functions/search.rs
@@ -151,6 +151,9 @@ pub struct SessionRegistry {
 struct SessionRecord {
     fingerprint: String,
     delivered: HashSet<String>,
+    /// Multi-KB guidance blocks this session already received; later
+    /// results carry a one-line pointer instead. Cleared with `delivered`.
+    preambles: HashSet<&'static str>,
     hint: Option<HintRecord>,
 }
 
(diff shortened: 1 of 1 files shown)


andersonleal:
**CI: harness-integration INT-005 failure was flaky, rerun is green.**

The failing run's artifacts show all turns terminal by 20:57:23 (parent ×2, leaf) and the failure in the *collect* phase: `trace-store change exceeded its deadline; traces never stabilized` — while iii-directory was busy downloading the MiniLM bundle. `iii-directory.out` has zero `search_functions`/`skills` calls in that run, so the guidance change could not be involved; INT-005 passed on the sibling PRs' runs. Re-ran the failed job: https://github.com/iii-hq/workers/actions/runs/34276437589 — success.

https://claude.ai/code/session_01PvtXK7JgHHNrG192afbRAk

PR #1123, opened 2026-09-09
perf(harness): reduce system prompt overhead

ytallo:
## Summary

- compact the default Harness system prompt while preserving its discovery, safety, trigger, Compose, sub-agent, and worker-authoring contracts
- add explicit efficiency guidance to avoid redundant reads, tests, validations, cosmetic edits, and repeated successful work
- limit each model response to at most one call with a large or multiline payload to reduce incomplete tool-call arguments
- add a regression test that preserves the efficiency guidance and caps the default prompt at 26 KB

## Motivation

A Harness regression comparison showed that the newer stack used 18.3% more turns and 41.4% more total tokens across comparable successful scenarios. Approximately 95% of the additional tokens were cache reads, indicating that extra turns repeatedly processed a larger accumulated context.

Trace analysis found that the increase was concentrated in `git_regression_forensics` and `shell_coder_sandbox`. Several large tool calls ended with incomplete arguments and had to be re-emitted; additional redundant validation and cosmetic follow-up edits further increased turns. The `minimal_path` control case also showed a fixed context increase with the same two-turn, one-call behavior.

This change addresses both multipliers: it reduces the fixed prompt footprint and adds behavioral guidance intended to avoid the extra turns observed in those traces.

## Prompt size

- bytes: 31,396 -> 22,552 (-28.2%)
- lines: 495 -> 364 (-26.5%)

## Validation

- `cargo test --lib prompt`
- `cargo test --test prompts`
- `cargo fmt -- --check`
- `git diff --check`

All 57 prompt unit tests and all 5 prompt integration tests pass.

## Summary by CodeRabbit

- **Improvements**
  - Streamlined assistant guidance for clearer, more efficient task execution.
  - Added direction to minimize unnecessary turns and calls, batch independent requests, handle multiline content reliably, and avoid repeating completed work.
  - Clarified delegation, asynchronous workflows, setup, language handling, and building new capabilities.
  - Simplified post-install verification guidance.

- **Reliability**
  - Added coverage to ensure efficiency guidance remains present and default instructions stay within supported size limits.

### DIFF
--- a/harness/prompts/default.txt
+++ b/harness/prompts/default.txt
@@ -1,48 +1,46 @@
 You are an iii agent worker.
 
-Use the language of the user's latest task message for all user-facing text: progress,
-tool descriptions, and the final response, unless the user explicitly requests another language.
-Keep that language across tool results and event notifications. Search capabilities stay in English.
+Use the user's latest task language for all user-facing progress, tool descriptions, event
+notifications, and final text unless another language is requested. Search capabilities stay in English.
 
-You have exactly one tool: `agent_trigger`. It calls a function on the iii engine. It takes
-three arguments: `function` (the function id, like `engine::functions::list`), `description`
-(a short user-facing description of the action in the language of the user's message), and `payload` (a JSON
-OBJECT with the function's arguments). Everything you do happens through `agent_trigger`.
-Never use a function id from memory.
+You have exactly one tool: `agent_trigger { function, description, payload }`. It calls a function
+through the iii engine; `description` is a short user-facing action label and `payload` is a JSON
+OBJECT. Never use a function id from memory.
 
-iii is a language agnostic runtime where services, agents, and tools are composed of the same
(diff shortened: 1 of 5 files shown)


sergiofilhowz (changes requested):
we should also update the iii and iii-minimal agent profiles

ytallo:
Addressed in 5b077977f: aligned the compact default prompt with the `iii` profile, added the same efficiency discipline to `iii-minimal`, and added regression coverage for both bundled profiles.\n\nValidated with `cargo test --manifest-path iii-directory/Cargo.toml --lib bundled_prompts_parse_and_carry_the_minimal_identity`, `cargo test --manifest-path iii-directory/Cargo.toml --lib bundled_iii_agent_body_is_the_harness_default_prompt`, and `cargo test --manifest-path harness/Cargo.toml prompt`.

sergiofilhowz (approved):

PR #1124, opened 2026-09-09
(MOT-4712) fix(console): serve on the next free port when the configured one is in use

anthonyiscoding:
## Summary
- Another console (or anything else) on the configured `http_port` stopped the worker with `binding TCP listener on 0.0.0.0:3113: Address already in use`.
- The port is now probed after the runtime configuration resolves and before the listener binds. When it is taken, the first free port above it is used.
- A warning names the configured and chosen ports. The port cell and the `console ready` line report the real port.

Needed for the `iii project init --learn-iii` tour (iii-hq/iii), which starts a second harness next to a running one. The harness template must pin a console release that contains this change.

## Test plan
- [x] `cargo check`, `cargo test --lib server`
- [ ] Start two consoles on the same host, confirm the second logs the warning and serves on 3114

## Summary by CodeRabbit

* **Bug Fixes**
  * The console now detects when the configured HTTP port is unavailable and automatically selects the next available port.
  * Startup logs show both the configured port and the port ultimately used.

### DIFF
--- a/console/src/main.rs
+++ b/console/src/main.rs
@@ -133,6 +133,15 @@ async fn main() -> Result<()> {
         }
     };
     cfg.http_port = runtime_config.http_port;
+    let free_port = server::free_port_from(cfg.http_port).await;
+    if free_port != cfg.http_port {
+        tracing::warn!(
+            configured = cfg.http_port,
+            chosen = free_port,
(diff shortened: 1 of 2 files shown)


anthonyiscoding:
Going back and forth on this one, somewhat useless until we can report from compose, closing from now

PR #1139, opened 2026-09-11
(MOT-4750) feat(workers-dev): a client of the iii compose daemon

andersonleal:
`workers-dev` stops being a supervisor of its own and becomes a TUI/CLI client of the `iii compose` daemon. **−6799 / +3197**, most of it deletion: the `cargo run` supervisor, the worker discovery, the dependency graph, the named stacks and the YAML writer are gone.

It was also broken at HEAD — `config.rs` required a root `worker-compose.yaml` to recognise the repo, and MOT-4614 moved that catalog to `.deploy/workers.yaml` without the `registry.dependencies` `discover.rs` needs. No invocation worked.

```
workers-dev                    # brings the stack up (engine included) and opens the dashboard
workers-dev restart harness    # the loop compose alone does not have
workers-dev start database     # a repo worker nothing declares yet: start it on demand
workers-dev stop               # compose::stop — containers, daemon, managed engine
```

```
workers-dev  ⎇ feat/workers-dev-compose-client  ws://127.0.0.1:49134  ns:my-project   ●4  ○9   up 4/13   daemon:3088448
┌ Workers 3/13 ────────────────────────────────┐ ┌ logs: compose (daemon)  ▶ live ──────────┐
│Worker                State    PID     UI     │ │→ session-manager starting                │
│ ▣ compose (daemon)   serving  3088448 —      │ │✓ session-manager ready (541ms)           │
│── stack (13) ──                              │ │✓ iii-directory ready (700ms)             │
│ ◐ queue              queued   —       —      │ │→ harness starting                        │
│ ● provider-anthropic ready    3540768 —      │ │                                          │
│ ◒ harness            starting —       ui     │ │                                          │
│── repo (60) ──                               │ │                                          │
│   database           —        —       ui     │ │                                          │
│   scrapling          registry —       —      │ │                                          │
└──────────────────────────────────────────────┘ └──────────────────────────────────────────┘
```

## What it took to get right

Four things the daemon does not hand a client, each found by a probe or a review rather than by reading the docs:

- **The first `--up` runs under a registered operation**, so the `compose-operation` trigger is the only live signal during a cold build. `compose::status` takes `try_lock` and answers from a `state.json` written *before* the operation began — without the feed, thirteen containers read `stopped` for the several minutes cargo spends compiling them. A trigger-initiated `up`/`restart`, by contrast, passes an unregistered uuid and publishes nothing.
- **`up`/`down`/`restart` are mutations and are not registered until that initial `--up` returns.** They answer `function_not_found` for the whole window; that is now a sentence, not a raw code.
- **`--engine` is a flag of `iii compose`, not of `iii`.** Passed before the subcommand the child exits 2 — which is the entire attach path, the one that survives an unclean death with an orphan engine still holding the port. Caught by review; my own live test only exercised the managed path.
- **A child's environment is a fixed baseline plus what the file declares**, so a key you exported reaches nothing. Every `*_API_KEY` in the launching shell is now forwarded through the env file the containers already read (mode `0600`). A provider's key belongs to `llm-router` — the provider only declares the variable's name; the router is the process that calls `std::env::var` on it.

## Starting the rest of the repo

The dashboard's second group is every worker with an `iii.worker.yaml` that the stack does not declare. `s` on one writes it a compose project **of its own** under gitignored `harness/.workers-dev/<name>.yaml` and starts it.

One file per worker, not one shared file: a container added to a project the daemon has already loaded stays invisible until a whole-project `compose::restart`, which stops everything else in it. A new file is a new project, so starting the fifth on-demand worker leaves the other four running. (`compose::add` would have spliced into a shared file and reconciled just the new container, but this build rejects its container-object form with `invalid type: map, expected a string`, and the string form cannot carry the `cargo run --bin` that 59 of the 73 repo manifests do not declare.) A pre-existing shared file is migrated on startup, without stopping what it is running.

## Changes outside the crate

- `harness/worker-compose.yaml`: an interpolated engine port, so a second worktree runs its own engine without forking the file, and one `env_file` entry per container. Both resolve to today's values when the variables are unset, so CI sees no change. **This puts `harness` in the changed-worker set and runs the full Harness Integration suite** — worth landing when main's is green.
- `harness/workers-dev.env.example`: tracked, empty, and load-bearing — compose validates every declared `env_file` when it first loads a project, so deleting it breaks every compose call against this one.
- `harness/DEVELOPMENT.md`: `iii compose up` is not a subcommand; it is `--up -f`. Plus how to bounce one worker.
- `harness/.gitignore`: the on-demand projects, and the skills `iii-directory` downloads at boot (they dirty `git status` for anyone who runs the stack).
- `harness/engine.config.yaml`, `state/src/main.rs`: two comments that still named workers-dev as the thing that starts and signals workers.

`harness/engine.config.yaml` is **not** deleted — `harness/Makefile` still reads it for the tmux stack.

## Verification

`cargo fmt`, `clippy --all-features -D warnings`, 35 inline tests. No CI job covers this crate; it is absent from `.deploy/workers.yaml` by design.

Live, against the real 13-container stack and a scratch project: managed spawn, attach to an existing engine (which survives `compose::stop`), single-container restart, the progress overlay through a cold build, `w` writing `III_STATE_UI_WATCH=1` and bouncing only that container, `w` refused with a message on a daemon we did not spawn, the `pnpm watch` sidecar cleaned up on exit, a forwarded `DEEPSEEK_API_KEY` read out of `/proc/<pid>/environ`, two workers started on demand without disturbing each other, and full teardown leaving no daemon, engine or sidecar behind.

https://claude.ai/code/session_018pmhe3CTYRzjuBTGNGkuXp

## Summary by CodeRabbit

- **New Features**
  - Workers Dev now provides a dashboard and remote control for Compose-managed projects.
  - Added container-level start, stop, restart, and status controls.
  - Added namespace selection, external worker directories, and on-demand worker projects.
  - Compose supports configurable engine ports and shared environment files.

- **Documentation**
  - Updated setup, startup, restart, troubleshooting, and multi-worktree guidance for the Compose workflow.

- **Changes**
  - Legacy stack configuration management and the logs command are no longer supported.


ytallo (approved):

andersonleal:
Follow-up commit: `--worker-dir` offers workers that do not live in this repo.

```bash
workers-dev --worker-dir ~/project/iii/harness-e2e
export WORKERS_DEV_WORKER_DIRS=~/project/iii/harness-e2e   # or once, in a shell profile
```

A directory that carries an `iii.worker.yaml` is one worker (the `harness-e2e` shape); otherwise its children are scanned, so another monorepo's root works too. The repo is read first, so a name it already uses is not replaced by a stranger. No auto-scan of siblings — a checkout with ten worktrees beside it would offer seventy copies of every worker.

Two things this turned up:

- **A worker that ships its own `worker-compose.yaml` is now taken at its word.** Its container declaration is used as written, with only `worker:` repointed at the real directory, `start_after` dropped (the containers it names are not in this project) and the env file appended. `harness-e2e` exits with `III_CONFIG is required` unless the `config_override` from its own file comes along, and a synthesized `cargo run --bin` declaration cannot guess it.
- **Starting an on-demand worker now always re-declares it.** `up` on a project the daemon has already loaded runs the declaration it loaded, not the one on disk — which is how the first attempt kept failing with a file that had since been fixed. A whole-project restart re-reads, and a project here holds exactly one container, so nothing else bounces.

Live: `harness-e2e` started from `/home/anderson/project/iii/harness-e2e` into the scratch stack's namespace, `ready` with its own config, listed in the `repo` group beside the repo's own workers.

PR #1212, opened 2026-09-21
fix: support engine 0.24.0 configuration initialization

guibeira:
## Summary

Restore worker startup compatibility with engines such as 0.24.0 that lack `configuration::ensure`, while retaining atomic initialization on engines that provide it.

- Cover 37 direct Rust initialization modules, the shared config client and its five additional consumers, and six Node workers (Claude Code, OpenCode, Pi, VS Code, Cursor, OpenWiki).
- Always attempt `configuration::ensure` first. Only its exact missing-function error enables legacy initialization; do not compare versions or cache capability detection.
- Legacy initialization reads the same entry with `raw: true` in `default`, then registers the original schema and metadata. Omit `initial_value` entirely for every existing non-null value; seed only for explicit null or exact uppercase `NOT_FOUND` from the lookup.
- Propagate malformed responses, lookup/register failures, transport errors, permission errors and bridge `ADAPTER_ERROR` without authorizing fallback writes.
- Preserve configuration identity, form metadata, seed normalization, routing, and intentional existing migrations.
- Warn once per entry/process that legacy initialization is non-atomic and recommend >=0.24.1 for concurrent-write safety.
- Update the configuration SOP and add a reproducible configuration-only validation runner.

## Important compatibility limitation

**Legacy initialization is not atomic.** Another writer can race the read/register pair, including cache-backed legacy registration. No client-side lock or extra read can provide the engine's atomic guarantee. The modern `ensure` initialization path remains unchanged and introduces no additional lookup/register calls in the helper.

## Validation performed

- 287 Node configuration/identity tests passed across all six workers.
- 14 shared Rust config-client tests passed, including real SDK error envelopes, strict retry wrappers, malformed responses, stored false/zero/empty-string values, templates, capability rechecks, and an explicit demonstration of the legacy race.
- ADE protocol fixture passed 14 scenarios using the real SDK against an isolated local TCP/WebSocket fixture; subprocesses run the test executable, never the `iii` CLI.
- All 42 Rust consumers compiled. Existing configuration tests passed where present; fp/web/kanban had zero tests matching the filter and provide compilation evidence only.
- Six permission tests passed: configuration mutators remain denied.
- Production TypeScript checks passed for OpenCode, VS Code and Cursor; Cursor test typechecking passed.
- Biome checks for the six production helpers and five TypeScript identity suites passed; changed Rust files passed rustfmt; `git diff --check` passed.

Runner: `scripts/check-configuration-compat.sh`.

## Pending before ready for review

- [x] Published engine 0.24.0 startup smoke test: the operator ran the isolated local test environment and reported PASS for version 0.24.0, all 14 declared workers ready using locally compiled binaries, and Console HTTP 200. The `--validate-only` run stopped the test environment afterward.
- [ ] Validate against a modern published engine endpoint.
- [ ] Exercise authenticated model calls and end-to-end workflows; startup validation does not establish full functional coverage or atomicity.
- [ ] Complete Claude Code/Pi full typechecking after their normal UI asset generation; current check is blocked by missing `ui-assets.generated.js`.
- [ ] Review the compatibility trade-off and shared Rust module inclusion across independently pinned workers.

No engine files, live configuration, dependency manifests/lockfiles, CI workflows or permission policies were changed. No release or merge is requested. Opening as **draft** for early inspection.

## Published-engine smoke test evidence

Operator-reported output from `/home/doggao/local_harness_test/run.sh --validate-only`:

```text
0.24.0
PASS: engine 0.24.0, 14 ready workers using local binaries, Console HTTP 200.
Legacy initialization remains NON-ATOMIC.
```

Local evidence: `/home/doggao/local_harness_test/test-logs/20260921-153810/validation.json`. This run was performed by the operator, not independently rerun by the agent. The machine-specific test directory and installer are outside this repository and are not included in this PR.

## Summary by CodeRabbit

* **New Features**
  * Configuration initialization now supports atomic setup with a compatibility fallback for legacy engines.
  * Existing configuration values are preserved during legacy initialization, while missing entries receive the default value.
  * Compatibility warnings are shown only once.

* **Bug Fixes**
  * Definitive missing-function and missing-entry errors now stop immediately instead of being retried.
  * Configuration failures are handled more consistently across workers and integrations.

* **Documentation**
  * Updated configuration guidance to describe modern and legacy initialization behavior.


guibeira:
@coderabbitai  review it.

